PR #8, opened 2022-11-28
Make minor changes to setup.*

arcondello:


setup.cfg
```diff
@@ -10,6 +10,7 @@ classifiers =
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
     Programming Language :: Python :: 3.10
+    Programming Language :: Python :: 3.11
```
thisac, line 13: 👍

setup.py
```diff
@@ -46,30 +46,20 @@ def build_extensions(self):
...
     name="dwave-gate",
     install_requires=[
         "numpy",
     ],
     ext_modules=cythonize(
-        extensions, annotate=bool(os.getenv("CYTHON_ANNOTATE", False))
+        ["dwave/gate/simulator/simulator.pyx",
+         "dwave/gate/registers/cyregister.pyx"
+         ],
+        annotate=bool(os.getenv("CYTHON_ANNOTATE", False))
```
thisac, line 58: Does this just make it a bit more compact, or does it change anything else?

thisac (approved):

setup.py
```diff
@@ -46,30 +46,20 @@ def build_extensions(self):
...
     ),
-    include_dirs=[numpy.get_include()],
+    include_dirs=[
+        numpy.get_include(),
+        'dwave/gate/simulator/'
+        ],
     cmdclass={"build_ext": build_ext_compiler_check},
-    zip_safe=False,
```
randomir, line 74: This package is not `zip_safe`, but I just checked, `zip_safe` is now obsolete. I guess we stopped recommending `python install ..` route?

arcondello: Oh, I removed it because it is specified in `setup.cfg`, not because it was obsolete

randomir: Got it. Damned GiHub diffs. :laughing:

setup.py
arcondello: Mostly more compact

arcondello:
I made another change to specify the include path in the file headers. That lets Cython have a better understanding of the relationship which is desirable when editing the files by hand

thisac (approved):
Looks good to me!

PR #12, opened 2022-12-20
Add support for measurements

thisac:
Adds support for:
* mid-circuit measurements
* measurement sampling in the simulator and on measurement objects
* expectation value calculations on measurement objects
* conditional operations based on previous measurements


thisac:
Will add some examples to the examples folder, as well as update the readme so that it takes these changes into account. Also haven't added a reno entry yet, so a few things are left before this is merge-ready.

README.rst
```diff
@@ -33,7 +33,7 @@ a two-qubit circuit.
 
     circuit = Circuit(2)
 
-    with circuit.context as q:
+    with circuit.context as (q, c):
```
randomir, line 36: API idea: introduce a `namedtuple` or `tuple` subclass like `Registers` (or `Context`?) and return that instead of a plain tuple from `CircuitContext.__enter__`.

That way users can be more or less verbose:
```
    ops.Hadamard(q[1])
    ...
```
or
```
with circuit.context as registers:
    ops.Hadamard(registers.q[1])
    ...
```

dwave/gate/simulator/simulator.pyx
```diff
@@ -41,7 +46,11 @@ def apply_instruction(
     little_endian: bool,
     conjugate_gate: bool = False,
 ):
-    if isinstance(op, ops.SWAP):
+    # if conditional and all bits are 1, skip
+    if op._cond and not any(op._cond):
+        return
```
randomir, line 51: Easier to parse (so no comment needed):
```suggestion
    if op._cond and all(op._cond):
        return
```

tests/test_operations/test_base.py
```diff
@@ -190,14 +196,61 @@ def test_matrix_repr(self, op, matrix):
...
 
 
-@pytest.mark.xfail(reason="Measurements are not implemented yet.")
 class TestMeasurement:
     """Unit tests for the ``Measurement`` class."""
 
+    def test_initialize_measurement(self):
+        """Test initializing a measurement operation."""
+        m = Measurement()
+        assert m.__repr__() == "<Measurement, qubits=None, measured=None>"
```
randomir, line 209:
```suggestion
        assert repr(m) == "<Measurement, qubits=None, measured=None>"
```

randomir (commented):
Very nice and clean code, pleasure to read. Just a few minor suggestions below.

README.rst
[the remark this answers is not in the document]
thisac: That's a quite nice idea actually! I'll think a bit about it (but will probably not make the change in this PR).

dwave/gate/operations/base.py
```diff
@@ -41,22 +41,34 @@
...
+from dwave.gate.registers.registers import ClassicalRegister, Variable
+from dwave.gate.tools.samples import sample
 from dwave.gate.tools.unitary import build_controlled_unitary, build_unitary
 
 if TYPE_CHECKING:
     from numpy.typing import NDArray
 
+# define an integer value for an arbitrary number of qubits;
+# used for clarity in variable qubit number operations
+ANY = -1
```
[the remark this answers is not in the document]
thisac: Also a good idea, but I'm not sure if it's necessary here. I'm quite OK with having -1 represent an arbitrary number of qubits, but I'll play around a bit with it.

> Or better yet to handle arithmetic comparison as well, perhaps raising (meaningful) error when ANY compared to number of qubits, you can implement a new type.

I think this could be the better option.

dwave/gate/primitives.py
```diff
@@ -69,31 +74,74 @@ class Bit:
...
 
+    def set(self, value: int, force: bool = False) -> None:
+        """Set a value for the bit.
+
+        Args:
+            value: Value that the variable should have.
+            force: Whether to replace any previously set value.
+        """
+        # leniently allow any type that evaluates to 0 or 1
+        value = int(value)
```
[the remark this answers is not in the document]
thisac: Hmm... this is I guess a conventional thing that the values should be either 0 or 1. I mean, booleans work equally as well, but I think I'd prefer having something like `value = int(bool(value))` in that case, so that the value type is kept as an integer for consistency.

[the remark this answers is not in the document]
randomir: Yes, since you're using that value later (missed that), `value = int(bool(value))` is actually what I had in mind -- i.e. accept a truthy, store as conventional bit.

dwave/gate/simulator/simulator.pyx
thisac: I'm realizing that both the code and the comment is wrong here. It should enter if at least one of the elements is `False`, so it should probably be `not all(op._cond)`. Thanks for highlighting that! 😅

tests/test_operations/test_base.py
thisac: I have a tendency to use `__repr__` instead of `repr` in test. It somehow _feels_ closer to testing the actual method. I'm not sure if there actually is a difference though, but I'd suspect not.

README.rst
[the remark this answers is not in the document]
arcondello: I want to second this idea. It will also make future changes easier because you won't need to, for example, update all of your test code like you did in this PR if you add another field. So it ends up keeping future changes more "contained".

tests/test_operations/test_base.py
arcondello: Though the user will be using `repr(...)` so that's what you want to be testing?

dwave/gate/simulator/simulator.pyx
randomir: Without actually checking what `op._cond` is :see_no_evil:, if this is a common condition, perhaps add a helper method to `Operation`?

dwave/gate/tools/samples.py
```diff
@@ -0,0 +1,33 @@
+import random
```
arcondello, line 1: missing header

tests/test_operations/test_base.py
randomir: Sounds reasonable, but `repr()` feels closer to user code :smile:

randomir: Damn GitHub, not refreshing comments like this :confused:

dwave/gate/simulator/simulator.pyx
thisac: Good point. I've added an `is_blocked` method [here](https://github.com/dwavesystems/dwave-gate/pull/12/files#diff-295c264653199148252d266a754a5553c61b034111d8504405adc2ad754ab60eR313-R322) instead.

dwave/gate/operations/base.py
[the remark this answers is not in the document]
thisac: @randomir Added an ``AnyClass`` to represent an arbitrary number of qubits instead ([see this commit](https://github.com/dwavesystems/dwave-gate/pull/12/commits/17802ea1b90c439de8813a73ebf4eaf1602b323e#diff-295c264653199148252d266a754a5553c61b034111d8504405adc2ad754ab60eR69-R105)). Let me know what you think. It inherits from int (to keep `_num_qubits` as ints`, while disabling arithmetics and comparisons.

dwave/gate/simulator/simulator.pyx
randomir: Perhaps make it a property (if you won't need future args)?

dwave/gate/primitives.py
```diff
@@ -69,31 +74,71 @@ class Bit:
...
+
+        Args:
+            value: Value that the variable should have.
+            force: Whether to replace any previously set value.
+        """
+        # leniently allow any type that evaluates to 0 or 1
+        value = int(bool(value))
+
+        if self._value and not force:
+            raise ValueError("Value already set. Use 'force=True' to replace it.")
```
randomir, line 128: This is not consistent with the docs above. Perhaps you meant:
```suggestion
        if self._value is not None and not force:
            raise ValueError("Value already set. Use 'force=True' to replace it.")
```

dwave/gate/operations/base.py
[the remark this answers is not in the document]
randomir: Seems like a total overkill, IMO. :smile: Also, I don't like `AnyClass` as `int` subclass (breaks Liskov).

I might not have the bigger picture here, but from the code, it looks like `ANY` will be used internally only (it won't seep outside object methods), so `None` would work just as well (special meaning, arithmetic/comparison disabled). OTOH, the same stands for the original `-1`, except that concrete "forbidden" int value with special meaning is too easy to mutate by accident (hence loosing the "special" meaning). If you need to keep `None` reserved for a different meaning (like value not assigned), then a simple `ANY = object()` is the way to go (creates "new `None`").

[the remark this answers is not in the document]
thisac: Fair enough. I might've gone a bit unnecessarily far on this one. 😅 I'm not too worried about Liskov in this case though, since it's more of one-off/single-instance used class, but I definitely get the point. The nice thing with using this class (or -1 for that matter) is that it stays an int, which is handy both for type-hinting and instance-checking. 

Personally, I have a slight preference to using either of those two alternatives, but I don't really have any strong feelings against just using a separate object placeholder (like `object()` or `...`) instead.

> it looks like ANY will be used internally only (it won't seep outside object methods)

That's correct. `None`, or even `...`, would definitely work, but maybe just not _as_ well.

dwave/gate/primitives.py
thisac: Yes, that's right, great catch!

dwave/gate/operations/base.py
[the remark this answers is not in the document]
randomir: I see type-hinting more as convenience than something driving the design. But if that's important here, you can create a new type (`Optional`-like), e.g. `Union[int, Any, None]`.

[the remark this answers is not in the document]
arcondello: Wouldn't it be more explicit to do
```python
class MyClass:
    def is_variable_length(self) -> bool:
        return True
    def num_qubits(self) -> int:
        raise TypeError("I don't have a fixed length")
```
that way you can get around a lot of these typing issues. Though of course it possibly ends up being too verbose.

README.rst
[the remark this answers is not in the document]
thisac: Fixed in #13.

dwave/gate/operations/base.py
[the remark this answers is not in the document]
thisac: I went with @arcondello's suggestion (more-or-less), and I reworked the way `_num_qubits`, `_num_control`, `_num_targets`, and `_num_params` are declared. This way, `None` represents "any" number of qubits and will raise the appropriate error if accessed before having been set when initialized.

dwave/gate/simulator/simulator.pyx
```diff
@@ -145,30 +153,106 @@ def simulate(
     state[0] = 1
 
     for op in circuit.circuit:
-        targets = [circuit.qubits.index(qb) for qb in op.qubits]
-        apply_instruction(num_qubits, state, op, targets, little_endian)
+        if isinstance(op, ops.Measurement):
+            _measure(op, state, circuit)
```
wbernoudy, line 157: Why treat `_measure` differently here? I think it should go with the rest of the instructions in `apply_instruction`.

dwave/gate/simulator/simulator.pyx
```diff
@@ -145,30 +153,106 @@ def simulate(
...
+            state_0_ids.append(i)
+        else:
+            weight_1 += np.abs(state[i])**2
+            state_1_ids.append(i)
+
+    sample = random.choices((0, 1), weights=[weight_0, weight_1])[0]
+
+    if collapse_state:
+        if sample == 0:
+            state[state_1_ids] = 0
```
wbernoudy, line 219: With large numbers of qubits this may be prohibitively slow. Better to do another loop here I think.

dwave/gate/simulator/simulator.pyx
```diff
@@ -145,30 +153,106 @@ def simulate(
...
+    if qubit >= num_qubits:
+        raise ValueError(f"Cannot sample qubit {qubit}. State has only {num_qubits} qubits.")
+
+    weight_0 = weight_1 = 0
+    zero_weight = False
+    state_0_ids = []
+    state_1_ids = []
+
+    for i in range(2**num_qubits):
+        if i % 2**(num_qubits - qubit - 1) == 0:
```
wbernoudy, line 206: This means only big-endian support, right? It should either support little-endian as well, or just raise an error in the simulate function.

dwave/gate/simulator/simulator.pyx
```diff
@@ -145,30 +153,106 @@ def simulate(
...
+            zero_weight = not zero_weight
+        if zero_weight:
```
wbernoudy, line 208: Something like `if (i >> (num_qubits - qubit - 1) & 1:` may be much faster here (and similar the `sample` function in the other file). Modulo on every loop is expensive

dwave/gate/simulator/simulator.pyx
thisac: Fair enough. I think the reason I put it here was to avoid having to loop over the target qubits a second time for measurments (in the `_measure` function), but it's a very small overhead.

arcondello (approved):

dwave/gate/simulator/simulator.pyx
thisac: Now raises a `CircuitError` if called with little-endian.

PR #13, opened 2022-12-22
Update context manager to return a `NamedTuple`

thisac:
Updates the context manager so that it returns a `NamedTuple` instead of a regular tuple. This allows for easy access to the registers without having to always unpack the tuple on context enter.

```python
with circuit.context as reg:
    X(reg.q[0])
    RX(0.5, reg.q[1])
```

works as well as

```python
    X(q[0])
    RX(0.5, q[1])
```

### DIFF
--- a/dwave/gate/circuit.py
+++ b/dwave/gate/circuit.py
@@ -23,7 +23,6 @@ __all__ = [
 ]
 
 import copy
-from dwave.gate.utils import cached_property
 from types import TracebackType
 from typing import (
     TYPE_CHECKING,
@@ -33,6 +32,7 @@ from typing import (
     Hashable,
     List,
     Mapping,
+    NamedTuple,
     Optional,
     Sequence,
     Tuple,
@@ -47,6 +47,7 @@ from dwave.gate.registers.registers import (
     QuantumRegister,
     SelfIncrementingRegister,
 )
+from dwave.gate.utils import cached_property
 
 if TYPE_CHECKING:
     from dwave.gate.operations.base import Operation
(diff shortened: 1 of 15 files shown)


releasenotes/notes/context-returns-namedtuple-83bda2334a7fb8d1.yaml
```diff
@@ -0,0 +1,4 @@
+---
+upgrade:
```
thisac, line 2: @arcondello Not sure if this should be under `upgrade`, or something else (e.g., `fix` or `other`).

dwave/gate/circuit.py
```diff
@@ -231,7 +232,14 @@ def num_parameters(self) -> int:
 
     @property
     def context(self) -> CircuitContext:
-        """Circuit context used to apply operations to the circuit."""
+        """Circuit context used to apply operations to the circuit.
+
+        A ``NamedTuple`` is returned when entering the context, which contains the quantum and
```
randomir, line 237: I'd say the
```
:class:`dwave.gate.circuit.Registers`
```
is returned, not just a generic `namedtuple`.

releasenotes/notes/context-returns-namedtuple-83bda2334a7fb8d1.yaml
randomir: Quacks like a backwards-incompatible new feature (but not because of the `namedtuple`, but because you now return two registers instead of one). So, you can put two entries, one under `features` and one under `upgrade`. But if you want to keep it to one note, the upgrade one is better.

Also, it's nice to include actual upgrade instructions here, like "instead of `with .. as q: q[0]` use `with .. as reg: reg.q[0]`".

randomir (approved):
LGTM

releasenotes/notes/context-returns-namedtuple-83bda2334a7fb8d1.yaml
thisac: Ah, good point. Thanks!

PR #18, opened 2023-01-05
Add documentation

thisac:
* Adds documentation to `dwave-gate`.
* Adds/updates module docstrings.


dwave/gate/mixedproperty.py
```diff
@@ -12,6 +12,11 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
+""":class:`mixedproperty` decorator.
```
randomir, line 15: I'm not sure we should promote this tiny utility function in the public docs. Alternatively, move the module to something like utils submod?

I like to treat **public docs as API boundary** -- what's not documented publicly it's considered internal. We are OK to break it without triggering a major release.

We never officially ratified this as our public policy, but we've been following it for the most part in Ocean.

dwave/gate/primitives.py
```diff
@@ -12,6 +12,10 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
+"""Primitive types used with quantum circuits.
+
+Contains primitive types such as qubits, bits and variables."""
```
randomir, line 17: nit: [`pep8`](https://peps.python.org/pep-0008/#documentation-strings) calls out (funnily enough) this as the most important bit on docstrings :laughing: :
> Note that most importantly, the """ that ends a multiline docstring should be on a line by itself

dwave/gate/mixedproperty.py
thisac: I agree with this sentiment. I'm actually not that fond of utils modules/files, and would actually prefer not to have a `utils.py` at all. Currently, it only contains a temporary patch for a Python-version restricted `cached_property` decorator.

I also agree with `mixedproperty` not necessarily having to be promoted as a public class, although I think I'm actually fine with that. Perhaps we can just have Sphinx ignore it and simply not render it instead?

dwave/gate/primitives.py
thisac: Ha, I recall this now as well! Wouldn't say it's _the_ most important thing though. 😅

docs/reference/operations.rst
```diff
@@ -0,0 +1,23 @@
...
+
+base module
+-----------
+
+.. automodule:: dwave.gate.operations.base
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
+operations module
```
JoelPasvolsky, line 17:
```suggestion
Operations Module
```

docs/reference/registers.rst
```diff
@@ -0,0 +1,15 @@
+Registers
+=========
+
+.. automodule:: dwave.gate.registers
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
+registers module
```
JoelPasvolsky, line 9:
```suggestion
Registers Module
```

docs/reference/simulator.rst
```diff
@@ -0,0 +1,23 @@
+Simulators
+==========
+
+.. automodule:: dwave.gate.simulator
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
+operation generation module
```
JoelPasvolsky, line 9:
```suggestion
Operation-Generation Module
```

docs/reference/simulator.rst
```diff
@@ -0,0 +1,23 @@
...
+---------------------------
+
+.. automodule:: dwave.gate.simulator.operation_generation
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
+simulator module
```
JoelPasvolsky, line 17:
```suggestion
Simulator Module
```

docs/reference/tools.rst
```diff
@@ -0,0 +1,23 @@
+Tools
+=====
+
+.. automodule:: dwave.gate.tools
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
+counters module
```
JoelPasvolsky, line 9:
```suggestion
Counters Module
```

docs/reference/tools.rst
```diff
@@ -0,0 +1,23 @@
...
+---------------
+
+.. automodule:: dwave.gate.tools.counters
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
+unitary module
```
JoelPasvolsky, line 17:
```suggestion
Unitary Module
```

docs/requirements.txt
```diff
@@ -0,0 +1,4 @@
+docutils==0.17.1
+sphinx==5.3.0
+sphinx-rtd-theme==1.1.1
```
JoelPasvolsky, line 3: Why not 'pydata_sphinx_theme' instead?

dwave/gate/circuit.py
```diff
@@ -12,6 +12,12 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
+"""Circuit and context manager classes.
+
+Contains classes to construct and handle quantum circuits which contain the quantum operations and
+instructions for running on simulators or hardware. See :py:mod:`dwave.gate.operations` for details
```
JoelPasvolsky, line 18: This is not an easy sentence to parse. The main part is clear, "Contains classes to construct and handle quantum circuits", but the subordinate clause to explain what a circuit is, "contain the quantum operations and instructions for running on simulators or hardware" leaves the verb "running" without a clear subject (running circuits or operations?).

dwave/gate/mixedproperty.py
```diff
@@ -12,6 +12,11 @@
...
+
+Contains a decorator for creating mixed properties, which differs from regular properties by
```
JoelPasvolsky, line 17: "creating mixed properties, which differs" --> the singular "differs" will take "decorator" for its subject but you mean the plural "properties".

dwave/gate/operations/operations.py
```diff
@@ -12,6 +12,12 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
+"""Quantum operations that can be applied in a circuit.
+
+Contains all supported quantum operations along with aliases for operations that may be know by
+different names, e.g., :class:`Fredkin` is an alias for :class:`CSWAP`, and :class:`Toffoli` and
```
JoelPasvolsky, line 18: "that may be known by" (typo: know --> known)
"different names; e.g." (pedantic, sorry)

dwave/gate/registers/__init__.py
```diff
@@ -12,4 +12,9 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
+"""Registers for collecting qubits, bits and variables.
+
+Contains performant container classes for keeping track of and handling primitive type object (see
```
JoelPasvolsky, line 17: "type object" --> "type objects"

dwave/gate/simulator/__init__.py
```diff
@@ -12,4 +12,9 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
+"""Simulators for running circuits locally.
+
+Contains a state-vector simulator able to return the resulting state vector after running a circuit
+on an all-zero initialized basis state. Support both the little- and big-endian conventions."""
```
JoelPasvolsky, line 18:
```suggestion
on an all-zero initialized basis state. Supports both the little- and big-endian conventions."""
```

dwave/gate/tools/__init__.py
```diff
@@ -12,5 +12,10 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
+"""Set of tools that are useful for circuit- and operation-creation.
+
+Contains a set of modules with helper functions for e.g., generation of different unitaries as well
```
JoelPasvolsky, line 17:
```suggestion
Contains a set of modules with helper functions for generation of different unitaries as well
```

dwave/gate/mixedproperty.py
randomir: Sure, whatever works for you, I don't have a strong opinion on how you implement it.

docs/conf.py
```diff
@@ -0,0 +1,41 @@
...
+
+autodoc_member_order = 'bysource'
+typehints_use_rtype = False  # avoids duplicate return types
+napoleon_use_rtype = False
+typehints_defaults = 'comma'
+
+# -- Options for HTML output -------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-html-output
+
+import sphinx_rtd_theme
```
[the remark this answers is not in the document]
thisac: Just used what seemed to be used in most other packages. I'm guessing it doesn't matter much for the local build anyhow. I have no preference towards using either one.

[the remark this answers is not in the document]
JoelPasvolsky: There's a slight advantage to being able to see how it will render in the actual public docs but I think this theme is quite a bit simpler in what is required to build 🤷

docs/index.rst
```diff
@@ -0,0 +1,45 @@
+.. _index_gate:
+
+dwave-gate
+==========
+
+.. include:: ../README.rst
+    :start-after: index-start-marker
+    :end-before: index-end-marker
+
+.. include:: readme
```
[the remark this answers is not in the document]
thisac: You're right, and I should probably change it here (since this file is only used for local builds).

docs/installation.rst
```diff
@@ -0,0 +1,6 @@
+Installation
+============
+
+.. include::  readme
```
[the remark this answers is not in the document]
thisac: This wouldn't render the installation text in the sdk-build though, I think, since it only uses a symlink to the docs folder. I could just make a copy of the readme to the docs folder, as I think is done in other packages.

docs/reference/circuit.rst
```diff
@@ -0,0 +1,23 @@
+Circuits
+========
+
+.. automodule:: dwave.gate.circuit
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
+mixedproperty module
+--------------------
```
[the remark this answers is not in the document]
thisac: Valid point. This is straightforward to fix, but I'm trying to think what would be best long-term since I've added module descriptions to the module `__init__.py` files, which won't be rendered if I remove the top-level `.. automodule::`. I'm guessing either

* Keep the module descriptions in the `__init__.py` and have the top-level `.. automodule::`, but add another heading for it (e.g., Circuit Module, for the example above) if it contains any methods, which most of them shouldn't.
* Move the module description to the rst files instead of having them in the  `__init__.py`.

Maybe I'm missing something here though, and there's a better way.

dwave/gate/circuit.py
thisac: I guess it can be sort of both, but what I mean here is running circuits on simulators/hardware. Would this suffice do you think? I can rewrite it otherwise, which might be a good thing to do.
```suggestion
instructions for running these circuits on simulators or hardware. See :py:mod:`dwave.gate.operations` for details
```

JoelPasvolsky: That solves the problem. Perhaps split into two: "Contains classes to construct and handle quantum circuits. The quantum circuits ..."?

JoelPasvolsky (approved):
LGTM. 
At some point I can go over the C docstrings too.

dwave/gate/registers/__init__.py
```diff
@@ -12,4 +12,10 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
+"""Registers for collecting qubits, bits and variables.
+
+Contains performant container classes for keeping track of and handling primitive type objects (see
```
thisac, line 17:
```suggestion
Provides performant container classes for keeping track of and handling primitive type objects (see
```

PR #19, opened 2023-01-26
Add copy of license to docs

thisac:
The SDK docs are failing (presumably) due to there being no license file in the docs folder.

* Add a copy of the license file to the docs folder.

### DIFF
--- a/docs/LICENSE
+++ b/docs/LICENSE
@@ -1,0 +1,203 @@
+                                 Apache License
+
+                           Version 2.0, January 2004
+
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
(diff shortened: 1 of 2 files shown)


arcondello:
Why not a symbolic link?

thisac:
I had that previously, but then it doesn't build locally on non-Unix systems. Not a huge issue, but this seems to be the solution in other packages so I went with this instead.

arcondello:
> I had that previously

It looks like you were just using a relative link in the .rst file, rather than having a symbolic link to the LICENSE (like say in the [cloud client](https://github.com/dwavesystems/dwave-cloud-client/blob/master/docs/LICENSE))

arcondello (approved):

thisac:
Oh, I mean previously as in when adding the docs to dwave-gate. Both the readme and license files were symbolic links, but didn't build correctly on Windows. Maybe this was just an issue with the readme though.

randomir:
For the record, git handles symbolic links, and so they work on all platforms `git` works on.

PR #20, opened 2023-01-27
Support for measurements in operation_generation.py

wbernoudy:
Fix some bugs in operation generation and add some features for supporting POVM measurements and Kraus operators. 

~Maybe will add some machinery for nicely supporting generating a cython/C function from an operator set which will "measure" each operator, and then "apply" a randomly sampled one.~ Done


dwave/gate/simulator/operation_generation.py
```diff
@@ -28,23 +28,34 @@ def binary(i: int, n: int) -> Tuple[int, ...]:
...
+            operation are ignored.
+
+        skip_zeros: If true, rows of given matrix that have no values will be ignored.
 
     Returns:
-        A tuple where the first item is the set of substates acted on by the gate, and the second
-        item is a dictionary mapping output substates to those corresponding instructions.
+        A tuple where the first item is the set of substates acted on by the gate, and
+        the second item is a dictionary mapping output substates to those corresponding
+        instructions.
```
thisac, line 49: Should the `output_sub_states` set be added to the description as well? Also, would it be more correct to say that the the dictionary is "mapping _dependent_ substates to those corresponding instructions", based on what they're named in the code?

dwave/gate/simulator/simulator.pyx
```diff
@@ -178,6 +185,52 @@ def _measure(op, state, targets):
         op._measured_qubit_indices.append(t)
 
 
+def _fast_sample(
```
thisac, line 188: 🐎💨

arcondello: I can only see this image as a horse farting and I am here for it.

randomir: My thoughts exactly.

dwave/gate/simulator/operation_generation.py
wbernoudy: Yes, good catch! Fixed

dwave/gate/simulator/simulator.pyx
```diff
@@ -168,7 +170,8 @@ def _measure(op, state, targets):
     op._measured_state = state.copy()
 
     for idx, t in enumerate(targets):
-        m = _sample(t, state)
+        # m = _sample(t, state)
+        m = _fast_sample(t, state)
```
thisac, line 174: I feel like we should just remove the old `_sample` function, and rename the `_fast_sample` to `_sample`, just replacing it. There's really no need to keep a slow sampling method anyway, right?

dwave/gate/simulator/simulator.pyx
```diff
@@ -178,6 +181,22 @@ def _measure(op, state, targets):
...
+    little_endian: bool = False,
+) -> int:
+    cdef int num_qubits = round(np.sqrt(state.shape[0]))
+
+    return measurement_computational_basis(
+        num_qubits,
+        state,
+        qubit,
+        little_endian=little_endian
+    )
```
thisac, line 197: Should `collapse_state` be passed to the `apply_operator` arg? Feel free to rename it as well to the latter, if you think that makes more sense, to keep it to a single name.
```suggestion
    return measurement_computational_basis(
        num_qubits,
        state,
        qubit,
        little_endian=little_endian,
        apply_operator=collapse_state,
    )
```

dwave/gate/simulator/operation_generation.py
```diff
@@ -466,6 +534,146 @@ def generate_op(
...
+    if not np.all(np.isclose(op_sum, np.eye(2**num_targets))):
+        raise ValueError("sum of operators by their conjugates must sum to identity")
+
+    c_codes = []
+
+    import itertools
+
+    full_op_options = {
+        (op_idx, little_endian, collect_norm):
+            f"{op_name}_op{op_idx}_{['norm', 'apply'][collect_norm]}_{['big', 'little'][little_endian]}_endian"
```
thisac, line 592: I think `'apply'` and `'norm'` are in the wrong order here, no?
```suggestion
            f"{op_name}_op{op_idx}_{['apply, 'norm'][collect_norm]}_{['big', 'little'][little_endian]}_endian"
```

dwave/gate/simulator/operation_generation.py
```diff
@@ -517,6 +742,24 @@ def generate_op(
...
+    op_set_definitions = [
+        ("measurement_computational_basis", 1, 0, dict(
+            operators=[measurement_0, measurement_1]
+        )),
+        ("dephase_noise", 1, 0, dict(
+            operators=[dephase0, dephase1]
+        )),
+        ("amplitude_dampening_noise", 1, 0, dict(
+            operators=[amp_damp0, amp_damp1]
+        )),
```
thisac, line 754: Is it correct to say that you've added these to the operations, but they are currently not used in any way by the simulator, i.e., there's no accessible support for them yet?

thisac (changes requested):
Looks great to me, apart from `apply` and `norm` seemingly being mixed up in the operation generation (although it seems to be happening both for the function names and when they're called, so I think the correct functions are actually called and everything is working).

dwave/gate/simulator/operation_generation.py
wbernoudy: Yes! Fixed

dwave/gate/simulator/operation_generation.py
wbernoudy: Yes that's right

dwave/gate/simulator/simulator.pyx
wbernoudy: Yes it should be. I think `collapse_state` is a good name for measurement, I just wanted to make it a bit more general sounding for the other stuff that the operator set functions allow.

Eventually this will be set to False when using `Measurement.sample()`, right?

thisac: > Eventually this will be set to False when using Measurement.sample(), right?

Yes, it should be set to `False`, although currently this sampling is implemented separately ([here](https://github.com/dwavesystems/dwave-gate/blob/master/dwave/gate/tools/samples.py#L22-L47)), so there's no collapsing taking place anyway.

tests/test_simulator/test_measurements.py
```diff
@@ -193,3 +193,37 @@ def test_conditional_op_multiple_qubits_true(self):
 
         assert np.allclose(res, expected)
         assert not x.is_blocked
+
+    def test_bell_state_measurement(self):
```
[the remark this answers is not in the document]
wbernoudy: How could you mock the measurement value?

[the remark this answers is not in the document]
thisac: You should probably be able to do something like:
```python
@pytest.mark.parametrize("bitval", [0, 1])
def test_bell_state_measurement(self, bitval, monkeypatch):
    ...
    monkeypatch.setattr(Bit, "value", bitval)
    ...
```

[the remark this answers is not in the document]
thisac: Oh, sorry, didn't think this one through. That wouldn't change the resulting state. 🤔

tests/test_simulator/test_measurements.py
```diff
@@ -245,7 +255,7 @@ def test_measurement_rng_seed(self):
         simulate(circuit, rng_seed=666)
         expected = tuple(b.value for b in circuit.bits)
 
-        for _ in range(100):
+        for _ in range(5):
             for bit in circuit.bits:
                 bit.reset()
             simulate(circuit, rng_seed=666)
```
thisac, line 261: Won't this always produce the same outcome since it resets the seed each time `simulate` is called?

thisac (approved):
Thanks @wbernoudy! Great update to measurements and the sampling!

tests/test_simulator/test_measurements.py
wbernoudy: It should, the idea behind the test is to ensure it with a high probability. With 10 qubits and 5 rounds, that comes out to a roughly 1 in 10^15 chance of it passing erroneously.

PR #21, opened 2023-01-30
Fix doctest failure

thisac:
Fixes failing doctest in the README.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -38,7 +38,7 @@ Example usage
 This example uses the ``dwave.gate.Circuit`` object's  context manager to append operations to
 a two-qubit circuit.
 
-.. code-block:: python
+.. testcode::
 
     import dwave.gate.operations as ops
     from dwave.gate import Circuit
@@ -46,6 +46,7 @@ a two-qubit circuit.
     circuit = Circuit(2)
 
     with circuit.context as (q, c):
+        ops.X(q[0])
         ops.Hadamard(q[1])
         ops.CZ(q[0], q[1])
         ops.Hadamard(q[1])
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -38,7 +38,7 @@ Example usage
 This example uses the ``dwave.gate.Circuit`` object's  context manager to append operations to
 a two-qubit circuit.
 
-.. code-block:: python
+.. testcode::
 
     import dwave.gate.operations as ops
     from dwave.gate import Circuit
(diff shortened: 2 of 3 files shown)


thisac:
I went with skipping it rather than fixing it, so that it passes. Otherwise the example would either have to be re-written as:

```python
>>> import dwave.gate.operations as ops
>>> from dwave.gate import Circuit
>>> circuit = Circuit(2)
>>> with circuit.context as (q, c):
...     ops.Hadamard(q[1])
...     ops.CZ(q[0], q[1])
...     ops.Hadamard(q[1])
<Operation: X, qubits=(<qubit: 0, id:...>,)>
<Operation: Hadamard, qubits=(<qubit: 1, id:...>,)>
<ControlledOperation: CZ, qubits=(<qubit: 0, id:...>, <qubit: 1, id:...>)>
<Operation: Hadamard, qubits=(<qubit: 1, id:...>,)>
```

which I think looks less clean, or some other workaround would have to be made (e.g., by adding a new block that is run by doctest but not rendered in the docs), which all felt hacky. This was the easy quick fix, but let me know if you have any better ideas.

JoelPasvolsky:
What about using `testcode` instead?

```
.. testcode::

    import dwave.gate.operations as ops
    from dwave.gate import Circuit

    circuit = Circuit(2)

        ops.Hadamard(q[1])
        ops.CZ(q[0], q[1])
        ops.Hadamard(q[1])

You can run the ``dwave.gate.simulator`` simulator on such circuits.

>>> from dwave.gate.simulator import simulate
>>> simulate(circuit)
```

I'm getting 

```
Failed example:
    simulate(circuit)
Expected:
    array([0.+0.j, 0.+0.j, 0.+0.j, 1.+0.j])
Got:
    array([1.+0.j, 0.+0.j, 0.+0.j, 0.+0.j])
```

which looks helpful

thisac:
Thanks @JoelPasvolsky. I'll try that! Ah, yes, I fixed that issue, but forgot to commit it and reverted it again before pushing. It's missing an X-gate (which I'll add). Is `testcode` a doctest-specific block?

JoelPasvolsky:
> Thanks @JoelPasvolsky. I'll try .... Is `testcode` a doctest-specific block?

Yes but it renders nicely in the HTML builder too: https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html

JoelPasvolsky (approved):

PR #22, opened 2023-02-01
Update makefiles and readme

thisac:
* Add a `make.bat` batch file for Windows.
* Add the pip installation step to the `make install` command.
* Update installation section in the readme.
* Add `make docs` command to makefile.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -15,8 +15,8 @@ dwave-gate
 
 .. index-start-marker
 
-``dwave-gate`` is a software package for constructing, modifying and running quantum circuits. It
-provides a set of tools that enables you to:
+``dwave-gate`` is a software package for constructing, modifying and running quantum circuits on the
+included simulator. It provides a set of tools that enables you to:
 
 * Construct quantum circuits using an intuitive context-manager interface.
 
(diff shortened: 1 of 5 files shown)


README.rst
```diff
@@ -38,7 +38,7 @@ Example usage
 This example uses the ``dwave.gate.Circuit`` object's  context manager to append operations to
 a two-qubit circuit.
 
-.. testcode::
+.. code-block:: python
```
thisac, line 41: @JoelPasvolsky This was changed _only_ in the readme in root, since the `testcode` block doesn't seem to render on GitHub. I've left it in the docs readme, since it's required for the doctests to pass, although it's not optimal to have two slightly different readme in root/docs.

randomir (commented):
LGTM, except that I don't understand why we need `make.bat` in the first place.

Why not require windows developers (if there are any) to `choco install make`, and then we don't need this duplication (in a horrible language that batch script is).

thisac:
> Why not require windows developers (if there are any) to choco install make, and then we don't need this duplication (in a horrible language that batch script is).

I mainly added it since it was easy and it would add immediate support for the make commands for Windows users (without having to install any third party package managers). We also already have a `make.bat` in the docs (due to Sphinx), and in many of our SDK docs folders, so it made sense just adding a connecting one in root.

Not a necessity, just a simple quality of life enhancement for a potential few.

arcondello (approved):

randomir (approved):

PR #23, opened 2023-02-01
Add method to retrieve qubits by label

### LINKED ISSUE #15: Consider adding a way to retrieve a qubit by label from a register
I would be nice to be able to do something like
```python
with circuit.context as register:
    q1 = register.bylabel("q1")
```

*Additional Considerations*
I believe that labels can be duplicated, so would need to handle that case somehow.

thisac:
* Adds a method `Circuit.get_qubit()` for retrieving a qubit by label (currently it's only possible to find a qubit index by passing the qubit object itself).
* Fixes bug where a measurement cannot be saved when measuring a qubit several times.

Fixes #15


arcondello (approved):
LGTM

PR #24, opened 2023-02-03
Fix various issues

### LINKED ISSUE #17: Inconsistent spacing when printing qubits
```python
In [1]: from dwave.gate import Qubit

In [2]: q = Qubit("a")

In [3]: str(q)
Out[3]: '<qubit: a, id:q0px>'
```
note the spacing between `qubit: a` and `id:q0px`.

Also, because labels can be arbitrary hashables, you might consider using `repr()` for the label, to disambiguate between `label=1` and `label='1'`

thisac:
* Check types when adding qubits/bits to a circuit and raise a ``TypeError`` if an incorrect type is passed to the method.
* Rename ``label`` to ``name`` for operations.
* Fix inconsistent spacing in primitive object representations. (#17)
* Fix sequential qubit/bit labelling when adding new registers.
* Update operation representations to use ``repr()`` for hashables. (#17)
* Adds `Circuit.get_bit` method (should've been in #23).
* Formats all files using black/isort (mainly affecting `operation_generation`).
* Ignore `operation_generation.py` in coverage since it's already indirectly tested in the simulator tests.

Fixes #17

### DIFF
--- a/dwave/gate/circuit.py
+++ b/dwave/gate/circuit.py
@@ -20,6 +20,7 @@ operations and instructions for running the circuits on simulators or hardware. 
 """
 
 from __future__ import annotations
+
 import itertools
 
 __all__ = [
@@ -396,7 +397,9 @@ class Circuit:
         qb, cb = len(self.qubits), len(self.bits)
         return f"<{self.__class__.__name__}: qubits={qb}, bits={cb}, ops={len(self.circuit)}>"
 
-    def get_qubit(self, label: str, qreg_label: Optional[str] = None, return_all: bool = False) -> Qubits:
+    def get_qubit(
+        self, label: str, qreg_label: Optional[str] = None, return_all: bool = False
+    ) -> Qubits:
         """Returns the Qubit(s) with a specific label.
 
         Args:
@@ -450,7 +453,9 @@ class Circuit:
 
         raise ValueError(f"Qubit {qubit} not found in any register.")
 
-    def get_bit(self, label: str, creg_label: Optional[str] = None, return_all: bool = False) -> Bits:
+    def get_bit(
+        self, label: str, creg_label: Optional[str] = None, return_all: bool = False
+    ) -> Bits:
         """Returns the Bit(s) with a specific label.
 
         Args:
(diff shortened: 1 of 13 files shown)


dwave/gate/circuit.py
```diff
@@ -355,7 +363,7 @@ def add_qregister(self, num_qubits: int = 0, label: Hashable = None) -> None:
         if label in self._qregisters:
             raise ValueError(f"Quantum register {label} already present in the circuit.")
 
-        data = [Qubit(str(i)) for i in range(num_qubits)]
+        data = [Qubit(str(i)) for i in range(self.num_qubits, self.num_qubits + num_qubits)]
```
arcondello, line 366: `QuantumRegister` only needs an iterable, not a sequence. So probably better to make `data` a generator rather than a list.

arcondello (approved):

dwave/gate/circuit.py
```diff
@@ -61,6 +62,7 @@
     from dwave.gate.operations.base import Operation
 
 Qubits = Union[Qubit, Sequence[Qubit]]
+Bits = Union[Bit, Sequence[Bit]]
```
randomir, line 65: I find this confusing. Imagine `list = element | list`.

You can't say `for bit in bits`, because it will sometimes fail and sometimes work.

thisac: I can definitely see that as confusing. A lot of places take either a (qu)bit or a sequence of (qu)bits, so this made things a bit cleaner, but I get your point. Wouldn't mind either renaming it e.g., `BitOrBits` or just using `Union[Bit, Sequence[Bit]]`, although I personally found this clear enough.

Part of me would have preferred just having either `Bit` _or_ `Sequence[Bit]` accepted, instead of in some cases supporting both. That might actually be the better solution long-term.

randomir: An elegant way would be to use a List monad (or sequence monad) for containing non-determinism of (qu)bit count here. (But that's not really an option in Python)

I guess the second best (for clarity) is to accept element _or_ sequence explicitly (single dispatch / pattern matching?).

But I definitely get the appeal of accepting `Union` and just dealing with it, so that user doesn't have to. I guess that's the "batteries included" approach, which is fine. It's just confusing if you have to deal with such a `Union` result.

PR #25, opened 2023-02-07
Update coverage config

thisac:
Coverage seemingly ignores the `omit` section in `.coveragerc` when running in CI. The issue only occurs when running Pytest via its CLI instead of as a Python module due to the omitted file paths being relative to root. This PR updates the omitted file paths so that they're found even if not run from root.

Note that running `python -m pytest` would have worked had it not been for the imports then using the local package (e.g., the relative import paths) instead of the installed package, thus failing to find certain files that have not been built locally.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -95,7 +95,7 @@ Tests and coverage can be run using Pytest.
 
 .. code-block:: bash
 
-    pytest tests/ --cov=dwave.gate
+    python -m pytest tests/ --cov=dwave.gate
 
 .. installation-end-marker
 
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -95,7 +95,7 @@ Tests and coverage can be run using Pytest.
 
 .. code-block:: bash
 
-    pytest tests/ --cov=dwave.gate
+    python -m pytest tests/ --cov=dwave.gate
```
thisac, line 98: Changing this so that it uses the same command as in the makefile. This is likelier to run the correct installation of pytest (if the user happens to have several versions), but with the downside (?) of testing the local files rather than the installed package.

PR #28, opened 2023-02-18
make dev build process work with code generation

juansebastianl:
At the moment building locally inplace doesn't work unless you first run ` dwave/gate/simulator/operation_generation.py`, this makes it so that the documentation is right and `python setup.py build_ext --inplace` should work. I don't think it's the most clean solution, but I couldn't get a solution with a custom `build_py` class as a command `python setup.py generate_gates` to work because of the `cythonize()` causing a build failure before the gates could be generated.


thisac:
Closing this due to staleness. Can bring it up again if we decide to update the build process.

PR #30, opened 2023-03-07
Store state in Circuit

thisac:
Changes the way that states are stored and accessed. Instead of being returned from the `simulate()` function, which now returns `None`, the state is stored in the `Circuit` object. This makes more sense since the measurement results are already stored in the `Circuit` object, inside the classical register bits, and will hopefully cause less confusion with the state only altering the circuit and not also returning results.

* The state is now stored in the ``Circuit`` object (same as the bits/measurement results) instead of being returned by the ``simulate()`` function. It can now be accessed via the `Circuit.state` property.

* A density matrix representation of the state can be accessed via the ``Circuit.density_matrix`` property, for both pure and mixed states (with the former being lazily calculated from the state vector).

### DIFF
--- a/dwave/gate/circuit.py
+++ b/dwave/gate/circuit.py
@@ -22,6 +22,13 @@ operations and instructions for running the circuits on simulators or hardware. 
 from __future__ import annotations
 
 import itertools
+import warnings
+from typing import TYPE_CHECKING
+
+import numpy as np
+
+if TYPE_CHECKING:
+    from numpy.typing import NDArray
 
 __all__ = [
     "CircuitError",
@@ -95,7 +102,10 @@ class Circuit:
         if num_bits is not None:
             self.add_cregister(num_bits=num_bits)
 
-        self._locked = False
+        self._state: NDArray = None
+        self._density_matrix: NDArray = None
+
+        self._locked: bool = False
 
     def __call__(self, qubits: Qubits) -> None:
         """Apply all the operations in the circuit within a circuit context.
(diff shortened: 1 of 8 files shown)


dwave/gate/circuit.py
```diff
@@ -220,6 +230,57 @@ def bits(self) -> ClassicalRegister:
 
         return bit_reg
 
+    @property
+    def state(self) -> Optional[NDArray]:
+        """The resulting state after simulating the circuit."""
+        if self._state is None and self._density_matrix is not None:
+            warnings.warn("State is mixed. Use 'Circuit.density_matrix' to access.")
```
arcondello, line 237: is this using the right stacklevel? I think you might want `stacklevel=2`

arcondello (approved):

dwave/gate/circuit.py
arcondello, line 237: Also, why a warning rather than an error?

thisac: Yes, you're right. Should be `stacklevel=2`.

thisac: Ah, I find it nicer not to raise an error if there is no state stored. If there's no state at all, it will just return `None`. In this case it will just raise a warning if the state only exists as a density matrix to notify the user that there is infact a state stored, but only as a density matrix (a mixed one).

arcondello: I guess I can think of two uses I might have as a user
a) I want to check whether the state is in the density matrix as part of my flow. In which case I want to do things like
```
if circuit.state is None:
    # do something else
```
in that case I don't think I want a warning.
b) It's always a mistake to access it in that state, in which case I would want an error.

PR #31, opened 2023-03-08
Add multi-qubit sampling to measurements

thisac:
Adds support for multi-qubit sampling in measurements along with support for directly returning samples as bitstrings instead of integers in nested lists. This simplifies sampling from simulating circuits and avoids the need for simulating the full circuit again for each sample.

```python
circuit = Circuit(2, 2)

    ops.Hadamard(q[0])
    ops.CNOT(q[0], q[1])
    m = ops.Measurement(q) | c

simulate(circuit)
```

Samples should (and will) always be either `"00"` or `"11"` for the specified circuit above. Which qubits to sample can be specified as e.g., `qubits=[0]` or `qubits=[0, 1]` (the latter sampling from all qubits in a 2-qubit circuit).

```pycon
>>> m.sample(num_samples=5, as_bitstring=True)
['00', '11', '00', '00', '11']
```

* `Measurement.sample()` and `Measurement.expval()` are updated to accept a sequence of qubit indices to sample, sampling all measured qubits if none are given.

* Measurement samples can be returned as bitstrings instead of integers in nested lists.


wbernoudy (approved):
LGTM!

PR #33, opened 2023-06-02
Update readme example

thisac:
Update readme example since the state is no longer returned when running the simulator (since #30).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -51,10 +51,14 @@ a two-qubit circuit.
         ops.CZ(q[0], q[1])
         ops.Hadamard(q[1])
 
-You can run the ``dwave.gate.simulator`` simulator on such circuits.
+You can run the ``dwave.gate.simulator`` simulator on such circuits,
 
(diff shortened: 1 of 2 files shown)


randomir (approved):
LGTM.

Btw, why is `docs/README.rst` not a symlink to the document in repo root?

thisac:
> Btw, why is `docs/README.rst` not a symlink to the document in repo root?

Basically to support building on Windows (see https://github.com/dwavesystems/dwave-gate/pull/18#discussion_r1063694827) and also the SDK build (thus no relative path to top-level readme).

randomir:
> Basically to support building on Windows (see [#18 (comment)](https://github.com/dwavesystems/dwave-gate/pull/18#discussion_r1063694827)) and also the SDK build (thus no relative path to top-level readme).

No, I mean literally git symlink (`docs/README.rst -> README.rst`). No need to keep duplicate files in the repo. See [an example](https://github.com/dwavesystems/dwave-cloud-client/tree/master/docs).

PR #35, opened 2023-06-06
Demos page

thisac:
Adds a demos page to the documentation, including a beginner's tutorial on how to use `dwave-gate` (taken from the examples folder). The demos are stored as Python notebooks in the `docs/demos/demos` folder and converted to Sphinx rst files and Jupyter notebooks using nbconvert and jupytext.

To convert a notebook into a valid Python notebook, run `jupytext --to py:sphinx name_of_notebook_file.ipynb`.

* Builds a page in the documentation with linked cards for the different demos.
* Stores the demos as Python notebooks for easier version control and reviews.
* Converts Python notebooks into Sphinx rst files for rendering.
* Converts Python notebooks into Jupyter notebooks for downloading.
* Runs every notebook and renders the outputs on each build.

**Note**: Requires Pandoc to build the correct files, along with Python packages `jupytext` and `nbconvert` (already in doc requirements), which might require some changes to the SDK build.


JoelPasvolsky (commented):
Just a couple of comments on installing dependencies to get the JN conversions working. It took me a while but mostly because I tried and moved on from Win and Py 3.7 environments before installing a pyenv env in my Ubuntu 16.04 VM and working with Py 3.9. 
Now that I have it working locally, I'll go through the rest.

docs/requirements.txt
```diff
@@ -2,3 +2,5 @@ docutils==0.17.1
 sphinx==5.3.0
 sphinx-rtd-theme==1.1.1
 sphinx_autodoc_typehints==1.19.5
+nbconvert==7.4.0
```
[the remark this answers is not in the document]
thisac: Ah, yeah, these requirements make sense. Thanks!

docs/requirements.txt
```diff
@@ -2,3 +2,5 @@ docutils==0.17.1
...
+jupytext==1.14.5
```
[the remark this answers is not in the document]
thisac: This should've been required previously as well (nothing to do with the demos setup), no?

docs/Makefile
```diff
@@ -8,10 +8,36 @@ SPHINXBUILD   ?= sphinx-build
...
+# 2. find all Jupyter notebooks and convert them into Sphinx rst files
+# 3. find all rst files and add the file name as a label at the top
+# 4. find all rst files and add the download links to the bottom
+.PHONY: demos
+demos:
+	mkdir -p $(OUTFOLDER)
+	cp -a $(DEMOFOLDER)/. $(OUTFOLDER)
+
+	find $(DEMOFOLDER) -name '*.py' -exec sh -c '$(JUPYTEXT) --to ipynb $$1 --output "$(OUTFOLDER)/$$(basename $$1 .py).ipynb"' sh {} \;
+	find $(OUTFOLDER) -name '*.ipynb' -exec $(NBCONVERT) {} --to rst --output-dir=$(OUTFOLDER) --execute \;
```
JoelPasvolsky, line 36:
```suggestion
	find $(OUTFOLDER) -name '*.ipynb' -exec $(NBCONVERT) {} --to rst --output-dir=$(OUTFOLDER) --execute --RegexRemovePreprocessor.patterns="^%" \;
```
**Changes:**
image
**to:**
image

I think that's preferable for our HTML documentation

thisac: Yes, thanks!

docs/Makefile
```diff
@@ -8,10 +8,36 @@ SPHINXBUILD   ?= sphinx-build
...
+	find $(OUTFOLDER) -name '*.rst' -exec sh -c 'sed -i "1s/^/.. _$$(basename $$1 .rst):\n\n/" $$1' sh {} \;
+
+	find $(OUTFOLDER) -name '*.rst' -exec sh -c 'sed -i -e $(ADD_DOWNLOAD_BUTTONS) $$1' sh {} \;
```
JoelPasvolsky, line 39: Very nice to have: additional postprocessing of ``*.rst`` files to replace major classes/methods with sphinx-build hyperlinks; for example `Circuit` --> :class:`~dwave.gate.circuit.Circuit`

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,134 @@
...
+# apply it to a circuit. Via the `operations` module, here shortened to `ops`, a variety of quantum
+# gates can be accessed.
+
+ops.X()
+
+###############################################################################
+# As we can see, any operation can be instantiated without declaring which qubits it should be
+# applied to. We can also call the matrix property on either the gate class itself or on an instance
+# (i.e., an instantiated operation, which can contain additional parameters and/or qubit
+# information).
```
JoelPasvolsky, line 41:
```suggestion
###############################################################################
# Notice above that an operation can be instantiated without declaring which qubits 
# it should be applied to. 
#
# You can also call the matrix property on either the gate class itself or on an 
# instance (i.e., an instantiated operation, which can contain additional 
# parameters and qubit information).
```

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,134 @@
...
+# Operations are applied when either the class, or an instance of the class, is called within a
+# circuit's context. They can be applied to the circuit in several different ways, as detailed
+# below. Both qubits and parameters can be passed either as single values (if supported by the gate)
+# or as sequences. Note that different types of gates accept slightly different argument, although
+# the qubits can _always_ be passed as sequences via the keyword argument `qubits`.
+#
+# When instantiated within a circuit context, operations must always be applied to specific qubits
+# which in turn must be part of the circuits qubit register. The qubit register can be accessed via
+# the named tuple returned by the context manager as `q`, indexing into it to retrieve the
+# corresponding qubit.
```
JoelPasvolsky, line 62:
```suggestion
# You apply operations by calling, within the context of a circuit, either a 
# class or an instance of a class. 
# 
# You can apply operations to a circuit in several different ways, as demonstrated
# below. You can pass both qubits and parameters as either single values (when 
# supported by the gate) or sequences. Note that different types of gates accept 
# slightly different arguments, although you can _always_ pass the qubits as 
# sequences via the keyword argument `qubits`.
#
# Always apply any operations you instantiate within a circuit context to 
# specific qubits in the circuit's qubit register. You can access the qubit 
# register via the named tuple returned by the context manager as `q`, indexing 
# into it to retrieve the corresponding qubit.
```

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,134 @@
...
+#
+# Let's start by creating a circuit object with 2 qubits in its register, and apply a single X-gate
+# to the first qubit.
```
JoelPasvolsky, line 65:
```suggestion
# This example starts by creating a circuit object with two qubits in its 
# register and applying a single X gate to the first qubit.
```

JoelPasvolsky (commented):
@thisac, I'm about halfway through the first demo, so I'll submit this half in case you have reservations, we can talk through before I make similar suggestions for the rest. It's too be more consistent with the writing style in the rest of our docs

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+for op in circuit.circuit:
+    print(op)
+
+###############################################################################
+# Note that e.g., CNOT and CX apply the exact same gate. CNOT is only an alias for CX (so they both
+# are labelled as CX in the circuit). There are also other aliases which you can spot either in the
+# source code for the operations module or read more about in the documentation.
+#
```
JoelPasvolsky, line 129:
```suggestion
# Note that the CNOT alias for the controlled NOT gate is labelled CX in the 
# circuit. You can find all operation aliases in the source code and documentation 
# for the operations module. 
```

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,256 @@
...
+###############################################################################
+# ## Applying gates to circuits
+#
+# You can apply operations to a circuit in several different ways, as demonstrated
+# in the example below. You can pass both qubits and parameters as either single
+# values (when supported by the gate) or sequences. Note that different types of
+# gates accept slightly different arguments, although you can _always_ pass the
+# qubits as sequences via the keyword argument `qubits`.
+#
+# :::note
```
JoelPasvolsky, line 115: Not working:
image

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,256 @@
...
+# Always apply any operations you instantiate within a circuit context to
+# specific qubits in the circuit's qubit register. You can access the qubit
+# register via the named tuple returned by the context manager as `q`, indexing
```
JoelPasvolsky, line 118:
```suggestion
# register via the named tuple, returned by the context manager as `q`, indexing
```

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,256 @@
...
+# into it to retrieve the corresponding qubit.
+# :::
```
JoelPasvolsky, line 120:
```suggestion
# 
```

image

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,256 @@
...
+###############################################################################
+# You can access all the operations in a circuit using the `Circuit.circuit`
+# property. The code below iterates over the returned list of all operations that
+# have been applied to the circuit.
+
+for op in circuit.circuit:
+    print(op)
+
+###############################################################################
+# :::note
```
JoelPasvolsky, line 157: Does not render:

image

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,256 @@
...
+# The CNOT alias for the controlled NOT gate is labelled CX in the circuit. You
+# can find all operation aliases in the source code and documentation for the
+# operations module.
+# :::
+
+###############################################################################
+# ## Simulating a circuit
```
JoelPasvolsky, line 164:
```suggestion
# ## Simulating a Circuit
```

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,256 @@
...
+#
+# `dwave-gate` comes with a performant state-vector simulator. It can be called by passing a circuit to the `simulate` method, which will update the quatum state stored in the circuit, accessible via `Circuit.state`.
```
JoelPasvolsky, line 166:
```suggestion
# `dwave-gate` comes with a performant state-vector simulator. You can call it by 
# passing a circuit to the `simulate` method, which will update the quantum state 
# stored in the circuit, accessible via `Circuit.state`.
```

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,256 @@
...
+
+from dwave.gate.simulator import simulate
+
+###############################################################################
+# We create a circuit object with 2 qubits and 1 bit in a quantum and classical registers respectively --- the bit is required to store a single qubit measurement --- and then apply a Hadamard gate and a CNOT gate to the circuit.
```
JoelPasvolsky, line 171:
```suggestion
# This example creates a circuit object with 2 qubits and 1 bit in a quantum and 
# a classical register respectively---the bit is required to store a single qubit 
# measurement---and then applies a Hadamard gate and a CNOT gate to the circuit.
```

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,256 @@
...
+
+circuit = Circuit(2, 1)
+
+with circuit.context as (q, c):
+    ops.Hadamard(q[0])
+    ops.CNOT(q[0], q[1])
+
+###############################################################################
+# We can now simulate the circuit, which will update its stored quantum state.
```
JoelPasvolsky, line 180:
```suggestion
# You can now simulate the circuit, updating its stored quantum state.
```

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,256 @@
...
+# 0, 0, 1\right]$ corresponding to the state:
+#
+# $$\vert\psi\rangle = \frac{1}{\sqrt{2}}\left(\vert00\rangle + \vert11\rangle\right)$$
+
+circuit.state
+
+###############################################################################
+# ## Measurements
+#
+# Measurements work like any other operation in dwave-gate. The main difference is that the operation generates a measurement value when simulated which can be stored in the classical register by piping it into a classical bit.
```
JoelPasvolsky, line 195:
```suggestion
# Measurements work like any other operation in `dwave-gate`. The main difference 
# is that the operation generates a measurement value when simulated, which can be 
# stored in the classical register by piping it into a classical bit.
```

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,256 @@
...
+#
+# We can reuse the circuit from above by simply unlocking it and appending a `Measurement` to it.
```
JoelPasvolsky, line 197:
```suggestion
# You can reuse the circuit above by simply unlocking it and appending a `Measurement` to it.
```

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,256 @@
...
+
+circuit.unlock()
+with circuit.context as (q, c):
+    m = ops.Measurement(q[1]) | c[0]
+
+###############################################################################
+# :::note
```
JoelPasvolsky, line 204: Not rendering

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,256 @@
...
+# We stored the measurement instance as `m`, which we can use for post-processing. It's also possible to do this with all other operations in the same way, allowing for multiple identical operation applications.
```
JoelPasvolsky, line 205:
```suggestion
# This example stored the measurement instance as `m`, which you can use for 
# post-processing. It's also possible to do this with all other operations in 
# the same way, allowing for multiple identical operation applications.
#
```

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,256 @@
...
+#     single_x_op = ops.X(q[0])
+#     # apply the X-gate again to the second qubit
+#     # using the previously stored operation
+#     single_x_op(q[1])
+# ```
+# This procedure can also be shortened into a single line for further convience.
+# ```python
+# ops.CNOT(q[0], q[1])(q[1], q[2])(q[2], q[3])
+# ```
+# :::
```
JoelPasvolsky, line 217:
```suggestion
# 
```

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,256 @@
...
+
+###############################################################################
+# The circuit should now contain 3 operations: a Hadamard, a CNOT and a measurment.
+
+print(circuit)
+
+###############################################################################
+# When simulating this circuit, the measurement will be applied and the measured value will be stored in the classical register. Since a measurement will affect the quantum state, the resulting state will have collapsed into the expected result dependent on the value which has been measured.
```
JoelPasvolsky, line 225:
```suggestion
# When you simulate this circuit, the measurement is applied and the measured 
# value is stored in the classical register. Since a measurement affects the 
# quantum state, the resulting state has collapsed into the expected result 
# dependent on the value which has been measured.
```

Another dangler: "When simulating this circuit, the measurement" has the measurement simulating the circuit

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,256 @@
...
+simulate(circuit)
+
+###############################################################################
+# If the measurement result is 0 the state should collapse into $\vert00\rangle$, and if the measurement result is 1 the state should collapse into $\vert11\rangle$. Outputting the measurement value and the state reveals that this is indeed the case.
+
+print(circuit.bits[0].value)
+print(circuit.state)
+
+###############################################################################
+# ## Measurement post-access
```
JoelPasvolsky, line 236:
```suggestion
# ## Measurement Post-Access 
```

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,256 @@
...
+
+print(circuit)
+
+###############################################################################
+# When simulating this circuit, the measurement will be applied and the measured value will be stored in the classical register. Since a measurement will affect the quantum state, the resulting state will have collapsed into the expected result dependent on the value which has been measured.
+
+simulate(circuit)
+
+###############################################################################
+# If the measurement result is 0 the state should collapse into $\vert00\rangle$, and if the measurement result is 1 the state should collapse into $\vert11\rangle$. Outputting the measurement value and the state reveals that this is indeed the case.
```
JoelPasvolsky, line 230:
```suggestion
# If the measurement result is 0 the state should collapse into $\vert00\rangle$, 
# and if the measurement result is 1 the state should collapse into $\vert11\rangle$. 
# Outputting the measurement value and the state reveals that this is indeed the case.
```

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,256 @@
...
+
+print(circuit.bits[0].value)
+print(circuit.state)
+
+###############################################################################
+# ## Measurement post-access
+# Since we stored the measurement operation in `m`, we can use it to access the state as it was before the measurement.
```
JoelPasvolsky, line 237:
```suggestion
# Since you stored the measurement operation in `m`, you can use it to access the 
# state as it was before the measurement.
```

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,256 @@
...
+#
+# :::note
```
JoelPasvolsky, line 239: Not rendering
image

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,256 @@
...
+# Accessing the state of the circuit along with any measurement post-sampling and state-access is only available for simulators.
+# :::
```
JoelPasvolsky, line 241:
```suggestion
# 
```

image

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,256 @@
...
+
+m.state
+
+###############################################################################
+# We can also sample that same state again using the `Measurement.sample` method, which by default only samples the state once. Here, we request 10 samples.
```
JoelPasvolsky, line 246:
```suggestion
# You can also sample that same state again using the `Measurement.sample` method, 
# which by default only samples the state once. Here, request 10 samples.
```

docs/demos/demos/intro_demo.py
```diff
@@ -0,0 +1,256 @@
...
+
+m.sample(num_samples=10)
+
+###############################################################################
+# Finally, we can calculate the expected value of the measurment based on a specific number of samples.
```
JoelPasvolsky, line 251:
```suggestion
# Finally, you can calculate the expected value of the measurement based on a 
# specific number of samples.
```

docs/demos/demos/intro_demo.py
thisac: Wonder why it seemingly works for me but not for you. How are you building the docs? :thinking:
image

JoelPasvolsky:
```
make demos
make html
```

thisac: Are you looking at the rendered docs (i.e., the rst file) or the jupyter notebook? The latter will not render the notebox, but the former should be displayed as above, rendering the docs nicely.

JoelPasvolsky: By "rendered docs" I mean the HTML. I get the same whether I use ``make html`` or I do "sphinx-build -b html`` after running ``make demos``.

image

If you don't see this too, I will try looking into it either tomorrow or early next week

docs/requirements.txt
```diff
@@ -2,3 +2,9 @@ docutils==0.17.1
 sphinx==5.3.0
 sphinx-rtd-theme==1.1.1
 sphinx_autodoc_typehints==1.19.5
+sphinx-design==0.4.1
+ipykernel==6.23.1
+matplotlib==3.71
```
JoelPasvolsky, line 7: I'm on an old Ubuntu with Py3.9, so this might not be supported anyway but I get 
```
ERROR: Could not find a version that satisfies the requirement matplotlib==3.71
```

thisac: Oh, should most definitely have been 3.7.1. Thanks for the discovery.

docs/Makefile
```diff
@@ -8,10 +8,54 @@ SPHINXBUILD   ?= sphinx-build
...
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
+.PHONY: demos
+demos:
+	# 0. create the output folder and copy all Python source files over to it
+	mkdir -p $(OUTFOLDER)
+	cp -a $(DEMOFOLDER)/. $(OUTFOLDER)
+
+	# 1. find all Python demos and convert them to Jupyter notebooks
+	find $(DEMOFOLDER) -name '*.py' -exec sh -c '$(JUPYTEXT) --to ipynb $$1 --output "$(OUTFOLDER)/$$(basename $$1 .py).ipynb"' sh {} \;
```
randomir, line 36: You can use shell's built-in [parameter expansion](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02) to avoid calling (external) `baseline`.

docs/Makefile
```diff
@@ -8,10 +8,54 @@ SPHINXBUILD   ?= sphinx-build
...
+		-e "s/\`\`operations\`\`/:mod:\`~dwave.gate.operations\`/g" \
+		-e "s/\`\`ops\.\(\w*\)\`\`/:class:\`~dwave.gate.operations.operations.\1\`/g" \
+		-e "s/\`\`Measurement\`\`/:mod:\`~dwave.gate.operations.base.Measurement\`/g" \
+		-e "s/\`\`ParametricOperation\`\`/:mod:\`~dwave.gate.operations.base.ParametricOperation\`/g" \
+		-e "s/\`\`ControlledOperation\`\`/:mod:\`~dwave.gate.operations.base.ControlledOperation\`/g" \
+		-e "s/\`\`ParametricControlledOperation\`\`/:mod:\`~dwave.gate.operations.base.ParametricControlledOperation\`/g" \
+		-e "s/\`\`simulate\`\`/:mod:\`~dwave.gate.simulator.simulate\`/g" \
+		-e "s/\`\`simulator\.\(\w*\)\`\`/:class:\`~dwave.gate.simulator.\1\`/g" \
+		-e $(ADD_DOWNLOAD_BUTTONS) \
+	$$1' sh {} \; \
```
randomir, line 57: If you prefer `sed`, I'd extract these substitution commands in a stand-alone sed script, just to avoid the escape hell.

docs/Makefile
```diff
@@ -8,10 +8,54 @@ SPHINXBUILD   ?= sphinx-build
...
+
+# Script to add file names as labels to the top of the rst files
+ADD_FILE_NAME_LABEL = "1s/^/.. _$$(basename $$1 .rst):\n\n/"
+
 # Put it first so that "make" without argument is like "make help".
 help:
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
+.PHONY: demos
+demos:
```
randomir, line 30: This build rule is sufficiently complex to warrant a stand-alone shell script.

Or, given how shell scripts are generally fragile, unmaintainable and unscalable, perhaps switch to something more portable and extensible, like Python? :slightly_smiling_face:

docs/Makefile
```diff
@@ -8,10 +8,54 @@ SPHINXBUILD   ?= sphinx-build
...
+	# 0. create the output folder and copy all Python source files over to it
+	mkdir -p $(OUTFOLDER)
+	cp -a $(DEMOFOLDER)/. $(OUTFOLDER)
+
+	# 1. find all Python demos and convert them to Jupyter notebooks
+	find $(DEMOFOLDER) -name '*.py' -exec sh -c '$(JUPYTEXT) --to ipynb $$1 --output "$(OUTFOLDER)/$$(basename $$1 .py).ipynb"' sh {} \;
+
+	# 2. find all Jupyter notebooks and convert them into Sphinx rst files
+	find $(OUTFOLDER) -name '*.ipynb' -exec $(NBCONVERT) {} --to rst --output-dir=$(OUTFOLDER) --execute --RegexRemovePreprocessor.patterns="^%" \;
```
randomir, line 39: Maybe you'll want to skip processing notebooks under `.ipynb_checkpoints/`.

docs/Makefile
```diff
@@ -8,10 +8,54 @@ SPHINXBUILD   ?= sphinx-build
...
+OUTFOLDER    = demos/out
+
+# Script to add download buttons to the bottom of the rst files
+ADD_DOWNLOAD_BUTTONS ="$\
+\$$a$\
+:download:\`Download Python source code \<$$(basename $$1 .rst).py\>\`\n\n$\
+:download:\`Download as Jupyter Notebook \<$$(basename $$1 .rst).ipynb\>\`\n"
+
+# Script to add file names as labels to the top of the rst files
+ADD_FILE_NAME_LABEL = "1s/^/.. _$$(basename $$1 .rst):\n\n/"
```
randomir, line 23: You can also use sed's **insert command** to prepend text (slightly simpler expression). When combined with **parameter expansion** switch as suggested below:
```suggestion
ADD_FILE_NAME_LABEL = "1i .. _$${1%.rst}:\n"
```

randomir (commented):
A few comments on the makefile.

docs/demos/demos/intro_demo.py
thisac: Strange. @JoelPasvolsky seemed to get it to work when building everything in Codespaces and suspected it had to do with his Ubuntu version. Since it seems to be causing issues I'll look into it, but as long as the demos are built correctly in the live docs I'd think it's fine.

JoelPasvolsky: I recently updated my Ubuntu so if it's needed to help debug I can run locally again and see if it now works

docs/Makefile
thisac: Thought about it, and it's a good idea. I don't necessarily mind the simplicity of having it like this though. Perhaps leaving it like this for the time being and potentially reevaluate later.

PR #40, opened 2023-07-18
Fix QIR load mid-function return

thisac:
Fixes a bug where a return statement in a block which is not the last block of the function is ignored. This fix causes a function to be terminated upon any encountering any `ret` instruction.


randomir (approved):
LGTM!

arcondello (approved):

PR #44, opened 2023-12-07
Unpin and update requirements

thisac:
* Unpins dev requirements
* Updates requirements for compatibility
* Adds pyqir as an optional dependency and removes the version check.


dwave/gate/qir/__init__.py
```diff
@@ -24,14 +24,14 @@
...
     pyqir_installed = False
 else:
-    # pyqir>=0.8 required
+    # pyqir>=0.9.0 required
     from importlib.metadata import version
 
     def parse(version: str) -> Tuple[str, str, str]:
         "Parse a MAJOR.MINOR.PATCH version string into a tuple."
-        return tuple(version.split("."))[:3]
+        return tuple(map(int, version.split(".")))[:3]
```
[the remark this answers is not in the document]
thisac: Yeah, good point. I didn't really think this through. 🤦‍♂️

dwave/gate/qir/__init__.py
```diff
@@ -24,14 +24,14 @@
...
 
-    pyqir_installed = parse(version("pyqir")) >= parse("0.8.0")
+    pyqir_installed = parse(version("pyqir")) == parse("0.9.0")
```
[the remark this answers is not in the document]
thisac: @randomir Decided to use an already existing version parser instead. Plan to loosen the pinned requirement to `>=0.9.0` but need to look into the build issue with 0.10.0 first.

dwave/gate/qir/__init__.py
```diff
@@ -24,14 +24,14 @@
...
+    from packaging.version import parse
 
-    def parse(version: str) -> Tuple[str, str, str]:
-        "Parse a MAJOR.MINOR.PATCH version string into a tuple."
-        return tuple(version.split("."))[:3]
+    pyqir_installed = True
 
-    pyqir_installed = parse(version("pyqir")) >= parse("0.8.0")
+    if parse(version("pyqir")) != parse("0.9.0"):  # pragma: no cover
+        raise ImportError(f"PyQIR v0.9.0 required, found {version('pyqir')}")
```
randomir, line 34: Maybe I'm missing something, but why isn't `pyqir` included as a dependency anywhere, not even in `extras`?

thisac: Not really, it's set as an optional dependency and, while it could (and maybe should) be part of extras, I wanted to check its compatibility before running any QIR compilation, e.g., in case `pyqir` is installed separately.

randomir: No, the check is fine, I'm just wondering how are users supposed to know they need to install `pyqir` manually? It's not in package deps, and it's not in readme (docs?).

thisac: Ah, I see. That's a good point. 🤔

thisac: @randomir Added PyQIR as an extra and also added a note in the installation instructions.

docs/reference/qir.rst
```diff
@@ -0,0 +1,31 @@
+QIR
+===
```
thisac, line 2: Had forgotten to add QIR to the docs.

pyproject.toml
```diff
@@ -34,6 +34,10 @@ requires-python = ">= 3.8"
 dependencies = [
     "numpy>=1.24.4",
 ]
+[project.optional-dependencies]
+    qir = [
```
arcondello, line 38: Should this be indented like this?

requirements.txt
```diff
@@ -1,4 +1,5 @@
+# working environment for Python>=3.8,<=3.11
 numpy==1.24.4
-cython==3.0.4
+cython==3.0.6
```
arcondello, line 3: if we're changing the Cython version, might as well use the latest

arcondello (approved):

pyproject.toml
thisac: Ah, probably not.

PR #50, opened 2024-06-25
Support NumPy 2.0 + Use M1 resource class in CircleCI

thisac:
* Update code to support NumPy 2.0.
* Remove `setup.cfg` in favour of using `pyproject.toml, and clean up the setup files.
* Use M1 resource class for macos builds on CircleCI.
* Add dependency tests for NumPy


pyproject.toml
```diff
@@ -3,10 +3,45 @@ requires = [
     "setuptools>=46.4.0",       # PEP-420 support, PEP-517/518 support, setup.cfg attr: support
     "wheel>=0.30.0",            # limited python api support
     "cython~=3.0",
-    'oldest-supported-numpy',
+    "numpy==1.24.4; python_version=='3.8'",
+    "numpy>=2.0.0rc1; python_version>='3.9'",
```
arcondello, line 7: I think it's fine to just do `numpy>=2.0.0`. The reason to do that is so `pip` doesn't pull in other `rc` in the future.

pyproject.toml
```diff
@@ -3,10 +3,45 @@ requires = [
...
 ]
 build-backend = "setuptools.build_meta"
 
+[project]
```
arcondello, line 11: I think it's fine either way, but out of curiosity why get rid of `setup.cfg`?

arcondello (commented):
IMO we should add dependency testing against NumPy in CI. See https://github.com/dwavesystems/dwave-preprocessing/blob/923620bfd259db3df5b83a7c367590a66a204946/.circleci/config.yml#L228-L255 for an example.

pyproject.toml
arcondello: I do see that [cibuildwheel](https://github.com/pypa/cibuildwheel) which is maintained by pypa has ditched their `setup.cfg`, so I am guessing that's what they want us to do. It's been quite a moving target over the last few years.

.circleci/config.yml
```diff
@@ -188,7 +194,7 @@ jobs:
           command: |
             python -m venv env
             . env/bin/activate
-            pip install dist/dwave-gate*.tar.gz
+            pip install dist/dwave_gate*.tar.gz
```
thisac, line 197: For some reason sdist started naming using and underscore instead of dash. I vaguely familiar with these being interchangeable with Python builds, but not sure about the details there.

arcondello: Yeah, I ran into the same thing in another PR :shrug:

pyproject.toml
thisac: Sure, using only `pyproject.toml` just seems to be the recommended way (see [this ](https://packaging.python.org/en/latest/guides/modernize-setup-py-project/)and [this](https://peps.python.org/pep-0621/)). I also think it's nicer keeping it simple, rather than having everything separated into several files.

arcondello: Sigh. Ok, I find [this](https://packaging.python.org/en/latest/guides/modernize-setup-py-project/) pretty definitive. I do wish they would pick a format and stick to it.

.circleci/config.yml
```diff
@@ -188,79 +194,92 @@ jobs:
...
+      - build-windows: *build
+      - test-codecov
+      - test-dependencies:
+          name: test-dependencies - << matrix.dependency-versions >> - py << matrix.python-version >>
+          requires:
+            - build-linux
           matrix:
             parameters:
+              # test the lowest and highest for each dependency
+              dependency-versions: [numpy==1.21.6, numpy]
```
thisac, line 251: @arcondello How important is it that we support `oldest-supported-numpy` vs. just setting a lower limit manually (in this case to support the typing module). I'd assume this is fine with regards to the SDK?

arcondello: Yes, that's fine. As long as you're testing the full range.

setup.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2022 D-Wave Systems Inc.
+# Copyright 2022-2024 D-Wave Systems Inc.
```
arcondello, line 1: Guidance from legal is just to leave it unless you totally overhaul the file. :shrug:

dwave/gate/simulator/simulator.pyx
```diff
@@ -26,9 +26,10 @@ import warnings
 from typing import List, Optional
 
 import numpy as np
-
 cimport numpy as np
 
+np.import_array()
```
arcondello, line 31: FWIW, I ended up not needing this for `dwave-samplers` and `dwave-preprocessing`. I am not 100% sure in what cases it's necessary or not.

arcondello (approved):

dwave/gate/simulator/simulator.pyx
thisac: Sure, not 100 if it's required either. I went with the recommendation from the [Cython docs](https://cython.readthedocs.io/en/latest/src/tutorial/numpy.html#adding-types):

> Therefore we recommend always calling "import_array" whenever you "cimport numpy"

arcondello: I am less convinced. If testing shows no issue then IMO better not to have a redundant call. It's just another surface for backwards compatibility. And in this code we're not using the NumPy API directly as far as I can tell.

thisac: I have no strong feelings, so I'm happy to remove it here if all seems to work without it.

pyproject.toml
```diff
@@ -3,10 +3,45 @@ requires = [
...
+    "numpy==1.24.4; python_version=='3.8'",
+    "numpy>=2.0.0; python_version>='3.9'",
 ]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "dwave-gate"
+dynamic = ["version"]
+authors = [
+    {name = "D-Wave Systems Inc.", email = "tools@dwavesys.com"},
```
arcondello, line 15:
```suggestion
    {name = "D-Wave Inc.", email = "tools@dwavesys.com"},
```

PR #53, opened 2024-10-04
Add support for Python 3.13 and remove 3.8

thisac:
* Adds support for Python 3.13
* Removes support for Python 3.8
* Minor cleanup, mostly typing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ commands:
     parameters:
       cibw-version:
         type: string
-        default: 2.16.2
+        default: 2.21.2
     steps:
       - run:
           name: run cibuildwheel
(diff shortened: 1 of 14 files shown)


.circleci/config.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         type: string
 
     docker:
-      - image: cimg/python:3.9
+      - image: cimg/python:3.12
```
arcondello, line 42: As @randomir points out in https://github.com/dwavesystems/dwave-cloud-client/pull/671#issue-2564499068, we could do standard Python images. I was contemplating switching over next time I touched these in `dwave-optimization` etc.

thisac: Guess I might as well do that here as well.

thisac: Got a `FileNotFoundError: [Errno 2] No such file or directory: 'docker'` error when switching to the `python:3.12` image, so kept the `cimg/python` one for now. Will look into, and update, later.

PR #59, opened 2026-01-28
Fix build warnings on invalid escape sequence

JoelPasvolsky:
The current patterns' strings' definitions 

```
else:
                pattern = "\%Qubit\* inttoptr \(i64 (\d+) to \%Qubit\*\)"
                qubit = re.search(pattern, str(arg)).groups()[0]
```

cause SDK [build warnings](https://app.circleci.com/pipelines/github/dwavesystems/dwave-ocean-sdk/3387/workflows/4a43a62c-07d8-4543-9a5c-399daf6722b9/jobs/48328) in newer Python versions:

```
<unknown>:176: SyntaxWarning: invalid escape sequence '\%'
<unknown>:184: SyntaxWarning: invalid escape sequence '\%'
```

I tested only by running your [unit test](https://github.com/dwavesystems/dwave-gate/blob/main/tests/test_qir/test_load.py) so you might want to test more.

Also updated sphinx requirements to match current SDK.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,5 +1,5 @@
-pydata-sphinx-theme==0.14.3
-sphinx==7.3.7
+pydata-sphinx-theme==0.16.1
+sphinx==8.2.3
 sphinx_autodoc_typehints==1.19.5
 
 reno[sphinx]==3.3.0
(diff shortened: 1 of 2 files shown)


thisac (approved):
LGTM

PR #60, opened 2026-06-05
Deprecate package

thisac:
Added deprecation notes in many places, incl. warning on import.


README.rst
```diff
@@ -1,3 +1,5 @@
+:warning: This version of *dwave-gate* is deprecated and no longer maintained
```
thisac, line 1: Not sure if there's a better message here to signal that not only this state of dwave-gate is deprecated. "This version" might not be the best way to refer to it.

PR #61, opened 2026-08-10
Update some classes used by QCDL devs (1st part)

JoelPasvolsky:
Transferred from another repo so stopped after two commits to ensure a clean git history before switching to work in this repo. 
@qci-amos, this one is short and should be checked more carefully. Thank you!

### DIFF
--- a/dwave/gate/qcdl/components.py
+++ b/dwave/gate/qcdl/components.py
@@ -1673,16 +1673,18 @@ class QCDLModule(QCDLModuleContainer):
     :class:`.Procedure` instance. This means that if a procedure is entered, a
     new :class:`.QCDLModule` is instantiated.
 
+    .. important:: This class is intended for use by developers of QCDL to
+        append instructions to a module, typically a qubit.
+
     This class is designed to provide an intuitive and convenient way to call
     the :meth:`~dwave.gate.qcdl.components.Procedure.add_statement` method. It does
     not perform validation and does not raise an error for unknown or invalid
-    methods and positional/keyword arguments. We recommend using the
-    :mod:`~dwave.gate.qcdl.operations` methods instead, where applicable, for
-    documentation and client-side validation.
+    methods and positional/keyword arguments. Use the
+    :mod:`~dwave.gate.qcdl.operations` methods instead, where applicable.
 
     Args:
-        module_name (str): Name of the module, typically a qubit.
-        proc (Procedure): :class:`.Procedure` instance this :class:`.QCDLModule`
+        module_name: Name of the module, typically a qubit.
+        proc: :class:`.Procedure` instance this :class:`.QCDLModule`
             is in.
 
     Examples:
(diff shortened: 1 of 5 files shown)


dwave/gate/qcdl/qcdl_models.py
```diff
@@ -400,7 +417,7 @@ def condition(self) -> Any:
             condition = self.args[-1]
         elif self.args:
             condition = self.args[0]
-        elif self.kwargs and "condition" in self.kwargs:
+        elifß self.kwargs and "condition" in self.kwargs:
```
thisac:
```suggestion
        elif self.kwargs and "condition" in self.kwargs:
```

dwave/gate/qcdl/components.py
```diff
@@ -1765,36 +1764,88 @@ def from_rewrapping(m: QcdlModule, new_proc: Procedure) -> QcdlModule:
     @property
     def qcdl_modules(self) -> tuple[QcdlModule]:
         """tuple[QcdlModule]: The :class:`~dwave.gate.qcdl.QcdlModule` this
```
thisac: Fine to remove this type, no?
```suggestion
        """The :class:`~dwave.gate.qcdl.QcdlModule` this
```

dwave/gate/qcdl/components.py
```diff
@@ -1765,36 +1764,88 @@ def from_rewrapping(m: QcdlModule, new_proc: Procedure) -> QcdlModule:
...
+
+                q0
+        """
         return self.qcdl_module_name
 
     @property
     def signal(self) -> str:
-        """str: Symbol representing this qubit's signal — a boolean that other
-        qubits may incorporate into their branch condition.
+        """str: Signal that other qubits can condition a branch upon.
```
thisac:
```suggestion
        """Signal that other qubits can condition a branch upon.
```

dwave/gate/qcdl/components.py
```diff
@@ -1765,36 +1764,88 @@ def from_rewrapping(m: QcdlModule, new_proc: Procedure) -> QcdlModule:
...
+
+            .. testoutput::
+
+                q0
+        """
         # use a tuple so that it's not editable
         return (self,)
 
     @property
     def name(self) -> str:
```
thisac, line 1800: Tangential to this PR, but why are there both a `name` and `qcdl_module_name` property? Should we only keep one? @qci-amos

thisac (approved):

dwave/gate/qcdl/components.py
qci-amos: fwiw, I previously needed to specify types like this for sphinx

dwave/gate/qcdl/components.py
```diff
@@ -1765,36 +1764,88 @@ def from_rewrapping(m: QcdlModule, new_proc: Procedure) -> QcdlModule:
     @property
     def qcdl_modules(self) -> tuple[QcdlModule]:
         """tuple[QcdlModule]: The :class:`~dwave.gate.qcdl.QcdlModule` this
-        container holds."""
+        container holds.
+
+        Examples:
+            This is an artificial example; see the example for the
+            :attr:`~dwave.gate.qcdl.QcdlModule.name` for comparison.
```
qci-amos: this will need to be updated wrt PR #64

dwave/gate/qcdl/components.py
qci-amos: I don't recall exactly, but we use `qcdl_module_name` in many places. I think the reason is because we have a `hasattr(v, "qcdl_module_name")` elsewhere in a context where things that aren't QcdlModule would have `name`, but then `name` is more natural to type out. I'm supportive of consolidating down to just one API method (and if needed, adding a `is_qcdl_module` method)

dwave/gate/qcdl/components.py
```diff
@@ -1852,15 +1926,20 @@ def get_other_qcdl_module(self, module_name: str) -> QcdlModule:
         Applying instructions to this will automatically register it as one of
         the modules used by this procedure.
 
+        .. todo:: I could not get a working example for this method
```
qci-amos, line 1933: I'd think this would work? 

```python
@qcdl()
def main(q0):
   q1 = q0.get_other_qcdl_module("q1")
```

dwave/gate/qcdl/qcdl_models.py
```diff
@@ -179,39 +181,56 @@ def __hash__(self) -> int:
...
+        *   -   :attr:`cargs`
+            -   Always ``None`` (Qiskit-style interface).
+        *   -   :attr:`condition`
+            -   Condition value for ``If``/``c_if``/``c_if_else``.
+        *   -   :meth:`simple_desc`
+            -   Short human-readable description.
+        *   -   :meth:`reassign_arg`
+            -   Replace a positional argument by index.
+        *   -   :meth:`reassign_kwarg`
+            -   Replace a keyword argument by name.
```
qci-amos: would it make more sense to just rely on docstrings in respective methods (and not document here)?

dwave/gate/qcdl/qcdl_models.py
```diff
@@ -532,14 +553,19 @@ class QcdlProcedureDef(BaseModel):
 
 
 class Qcdl(BaseModel):
-    """Pydantic model for the top-level QCDL compiler payload.
+    """`Pydantic <https://pydantic.dev/docs/>`_ model for the QCDL compiler.
+
+    .. important:: This class is intended for use by developers of QCDL.
```
qci-amos: I would say that the construction of the class is for developers, but the interfaces are more broadly useful

dwave/gate/qcdl/qcdl_models.py
```diff
@@ -532,14 +553,19 @@ class QcdlProcedureDef(BaseModel):
 
 
 class Qcdl(BaseModel):
-    """Pydantic model for the top-level QCDL compiler payload.
+    """`Pydantic <https://pydantic.dev/docs/>`_ model for the QCDL compiler.
```
qci-amos:
```suggestion
    """`Pydantic <https://pydantic.dev/docs/>`_ model for a QCDL program.
```

dwave/gate/qcdl/qcdl_models.py
```diff
@@ -532,14 +553,19 @@ class QcdlProcedureDef(BaseModel):
...
+
+    .. important:: This class is intended for use by developers of QCDL.
+
+    Use the :meth:`~dwave.gate.qcdl.qcdl_circuit.QcdlCircuit.to_model()` method
+    to generate an instance of this class.
```
qci-amos: it probably makes more sense to reference the `@qcdl` decorator than the `QcdiCircuit.to_model()` API.

dwave/gate/qcdl/qcdl_models.py
```diff
@@ -532,14 +553,19 @@ class QcdlProcedureDef(BaseModel):
...
 
-    :meth:`QcdlCircuit.to_model()` returns an instance of this class directly.
-    Use :meth:`model_dump` to recover a plain :class:`dict` suitable for the
-    compiler.
+    You can recover a plain :class:`dict` (compatible with the compiler) with
+    the :meth:`~pydantic.BaseModel.model_dump` method.
```
qci-amos:
```suggestion
```
I'm ok with skipping explanations for how pydantic works because our user facing interfaces will handle the conversion automatically.

qci-amos (approved):

dwave/gate/qcdl/components.py
JoelPasvolsky: @qci-amos, returns the same error I was seeing in all my failed attempts:

```
QCDLInternalError: QCDL Modules list has not been provided for <Procedure main>
```

PR #64, opened 2026-08-11
Capitalize qcdl in class names

thisac:
* Capitalize all instances of `Qcdl` in class names (e.g., `QcdlCircuit` to `QCDLCircuit`)
* Renamed the `Qcdl` class to `QCDLProgram`.

### DIFF
--- a/dwave/gate/qcdl/qcdl_models.py
+++ b/dwave/gate/qcdl/qcdl_models.py
@@ -18,14 +18,14 @@ This module provides Pydantic-based models for validating, documenting, and
 serializing QCDL payloads.  The models add runtime validation, field-level
 documentation, and structured serialization/deserialization.
 
-:class:`QcdlStatement` also surfaces a full derived API
+:class:`QCDLStatement` also surfaces a full derived API
 (``modules``, ``qargs``, ``condition``, ``simple_desc``, etc.) directly
 on the model for statement-level logic.
 
 Round-tripping
 --------------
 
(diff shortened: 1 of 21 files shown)


JoelPasvolsky:
@thisac, I didn't look through all of this, it looks consistent and you caught the doc xrefs too.

dwave/gate/qcdl/qcdl_models.py
```diff
@@ -502,39 +502,39 @@ class QcdlSignature(BaseModel):
...
         ),
     )
-    signature: QcdlSignature = Field(
+    signature: QCDLSignature = Field(
         description="Qubit and argument specification for calling this procedure.",
     )
 
 
-class Qcdl(BaseModel):
+class QCDLModel(BaseModel):
```
qci-amos: I'm not excited about this name change because in my mind its "model" association with BaseModel is an implementation detail. I think the reason you did this is to avoid the all-caps `QCDL`? I'd like to think if there's a better name. One idea is to call it `Program` since that matches what I've called the kwarg that receives this.

thisac: Yes, I at least wanted to start the discussion. I'm fine with either just `QCDL` or `QCDLProgram` (slight preference towards the latter).

qci-amos: the later sounds good to me

qci-amos (approved):

PR #67, opened 2026-08-14
remove QCDLNotImplementedError

### LINKED ISSUE #66: I think we can just remove QCDLNotImplementedError
https://github.com/dwavesystems/dwave-gate/blob/d3aa843a407c08ffdb7bb1ef9de00fcdd60f1d7f/dwave/gate/qcdl/exceptions.py#L28-L29

qci-amos:
closes #66 
I don't remember why it was added in the first place and in any case, it isn't used now (we should just use the built-in one on the client, we have a different exception for server-side "not implemented").


JoelPasvolsky (approved):

thisac (approved):
Thanks @qci-amos !

PR #72, opened 2026-08-19
QCDL operations: add reset and sxdg, and validate qubit arguments

qci-amos:
Part of splitting #71 into reviewable pieces. This one is the `operations`
module; the other four PRs are independent of it and of each other.

Two commits, reviewable separately.

### 1. Add `reset` and `sxdg` operations

`reset` already existed as `q0.reset()`, but only through `__getattr__`, so it
had no signature, no docstring, and no importable name — the user guide reached
for it in a way nothing else in the guide does. This adds the operation form.
It is an alias for the same statement, not a variant (there is a test asserting
the two produce identical programs).

Unlike `initialize`, which loops until every qubit in the program is reset,
`reset` acts on one qubit and has a deterministic duration, so it can be used
inside a conditional branch. The guide's erasure-detection example now uses it.

`sxdg` is the adjoint of `sx` and was simply missing from the gate set.

### 2. Validate the qubit arguments of every operation

Two failure modes were unpleasant:

- Passing something that is not a qubit surfaced much later as
  `AttributeError: 'int' object has no attribute 'procedure'`, naming an
  internal attribute rather than the argument at fault.
- A two-qubit gate given the same qubit twice — `cx(q0, q0)` — built cleanly
  and was rejected only once the program reached the service.

`_validate_qubit_args` wraps every operation and derives the rule from the
signature rather than from a hand-maintained list:

- each **named** `QCDLModule` parameter is a separate role in the operation, so
  two of them may not be the same module;
- a `*qubits` parameter is a *set* of qubits to act on, where a repeat is
  harmless (`barrier`, `initialize`).

A test asserts those two categories stay disjoint and cover every multi-qubit
operation, so a new operation lands in one of them rather than escaping the
check. Another asserts every operation is actually wrapped.

The wrapper preserves `__name__`, `__doc__`, `__module__` and the signature, so
autodoc is unaffected, and it defers to Python for arity errors so
`cx(q0)` still reports against the real signature.

Also adds `__all__`, so `from dwave.gate.qcdl.operations import *` brings in the
operations rather than this module's own imports (`np`, `inspect`,
`implementations`, ...).

### Compatibility

Code that passed a non-qubit, or the same qubit twice to a two-qubit gate, now
raises `QCDLUserError` at build time instead of failing later or being rejected
by the service. Nothing that built a valid program changes.

### Testing

`pytest tests/` passes. New tests are parameterized over every operation
discovered by introspection, so they cover gates added later too.


docs/workflow.rst
```diff
@@ -517,11 +517,11 @@ You can also reset qubits individually.
 .. testcode::
 
     from dwave.gate.qcdl import qcdl
-    from dwave.gate.qcdl.operations import initialize
+    from dwave.gate.qcdl.operations import reset
```
JoelPasvolsky, line 520: Good catch, thanks!

docs/workflow.rst
```diff
@@ -830,15 +830,15 @@ This example detects and resets a qubit if it has been erased.
 .. testcode::
 
     from dwave.gate.qcdl import qcdl
-    from dwave.gate.qcdl.operations import mced
+    from dwave.gate.qcdl.operations import mced, reset
 
     @qcdl(1)
-    def detect_erasure_example(q):
-        erased = q.Register(name="erased")
+    def detect_erasure_example(q0):
```
JoelPasvolsky, line 836: Another good catch

dwave/gate/qcdl/operations.py
```diff
@@ -88,6 +236,36 @@ def initialize(*qubits: QCDLModule) -> None:
     qubits[0].initialize(*qubits[1:])
 
 
+@_validate_qubit_args
+def reset(qubit: QCDLModule) -> None:
+    r"""Reset a single qubit to :math:`|0\rangle`.
+
+    Unlike :func:`initialize`, which loops until every qubit in the program is
+    reset, this operation acts on one qubit and has deterministic duration, so
+    it can be used inside a conditional branch.
```
JoelPasvolsky:
```suggestion
    Unlike :func:`initialize`, which ensures every qubit in the program is
    reset, this operation acts on one qubit. It has deterministic duration, so
    you can use it inside a conditional branch.
```

Removed some implementation details

JoelPasvolsky (approved):

dwave/gate/qcdl/operations.py
```diff
@@ -55,19 +55,186 @@ def qcdl_module_methods(q0, q1):
...
+def _is_qubit(value: Any) -> bool:
+    """Whether ``value`` can carry a QCDL instruction.
+
+    A machine may supply its own module type rather than a
+    :class:`~dwave.gate.qcdl.QCDLModule`, so this also accepts anything that
+    identifies itself as a QCDL module.
+    """
+    if isinstance(value, QCDLModule):
+        return True
+    return getattr(value, "_is_qcdl_module", False) is True
```
thisac: Are there any internal objects that have the `_is_qcdl_module` attr but are not `QCDLModule`? Is this otherwise documented as a requirement for "custom" QCDL modules?

dwave/gate/qcdl/operations.py
```diff
@@ -55,19 +55,186 @@ def qcdl_module_methods(q0, q1):
...
+    Without this, a non-qubit argument surfaces much later as an
+    ``AttributeError`` naming an internal attribute, and a two-qubit gate given
+    the same qubit twice builds cleanly and is only rejected by the service.
+
+    Whether the qubits have to be distinct follows from the signature: each
+    named qubit parameter is a separate role in the operation, so two of them
+    may not be the same module, whereas a ``*qubits`` parameter is a set of
+    qubits to act on, where a repeat is harmless.
+
+    A ``*qubits`` parameter also binds happily to nothing at all, so python
```
thisac:
```suggestion
    A ``*qubits`` parameter also binds happily to nothing at all, so Python
```

PR #73, opened 2026-08-19
QCDL registers: report a register name that is allocated twice

qci-amos:

### The problem

The compiler keeps the **first** allocation of a register name, so a second
declaration of the same name on the same module is a no-op — its initial value
never reaches the qubit. That was silent, and it is almost always a mistake.

The signal example in the user guide was itself relying on it:

```python
send_register = sc.Register(name=name)          # allocates on q0 and q1
measure(q0, register=q0.Register(name=name))    # silently discarded
```

### The change

Allocated register names are tracked on the circuit and a re-declaration
raises `QCDLUserError`, naming the dtype it already has, the procedure that
allocated it first, and the ways to say the reuse was deliberate.

**Scope.** Names are global to the circuit, not local to a procedure — a name
allocated in one procedure is the same qubit memory as that name in another —
so the record lives on `QCDLCircuit.allocated_registers` rather than on
`Procedure`. It is still keyed per module, so `q0` and `q1` may each hold a
register called `"r0"`.

**Procedure re-runs.** Calling a procedure re-runs its body while the program
is being built, even though the procedure is emitted once, so every call would
otherwise look like a re-declaration of whatever registers it declares. Each
run is a distinct `Procedure` instance sharing one `proc_name`, and `proc_name`
is what the circuit already deduplicates procedures on, so an allocation
reached through a different run of the same procedure is not treated as a
clash. A duplicate *within* one body still is.

**The two opt-outs** stay available, and differ in scope:

- **`alias=True`** is always an alias and never allocates, so an
  `initial_value` is rejected whatever the state of the name — already
  allocated or not.
- **`ignore_reallocation=True`** is a no-op only *once the name is allocated*.
  An `initial_value` is rejected only in that case; a name that is not yet
  allocated is allocated as usual and may carry one.

Where a value is rejected the reasoning is the same: opting in to the
re-declaration says the existing memory is what you want, whereas giving a
value says the opposite, and the compiler would ignore it either way.

To tell "no initial value given" apart from an explicit `0`, `initial_value`
now defaults to `None` and resolves to `0` or `0.0` according to the dtype.
`Register(0, name="dup", ignore_reallocation=True)` is therefore an error when
`dup` already exists, while `Register(name="dup", ignore_reallocation=True)` is
not — passing the default explicitly is still saying something.

The guide's signal example and the equivalent example in the `measure`
docstring now say `alias=True` where they mean to reuse the scope register.

### Compatibility

This is the one PR in the split with a real chance of breaking existing
programs, by design: any code that declared a name twice was silently losing
the second initial value and now gets an error that says so. That includes a
procedure that reuses a name its caller already took, which the global scoping
newly catches. The fix at each site is one of `alias=True`,
`ignore_reallocation=True`, a different name, or reusing the handle that
already exists.

### Testing

`pytest tests/` passes. `tests/test_registers.py` covers each opt-out on both a
fresh and an already-allocated name, the explicit-zero case, `Array` (which
always carries contents and so can never opt in), per-module scoping, the
global-across-procedures cases (clash, which procedure gets named, a procedure
called twice, a duplicate inside one body, and the same name on two qubits),
and the narrowing-with-an-alias pattern the guide now teaches.


dwave/gate/qcdl/components.py
```diff
@@ -76,6 +76,21 @@ def default(self, obj: Any) -> Any:
             return str(obj)
 
 
+class RegisterAllocation(NamedTuple):
+    """One entry in the register names a circuit has allocated.
```
JoelPasvolsky:
```suggestion
    """Registers by name an allocated circuit.
```

This is my guess, I am having difficulty parsing the current description

dwave/gate/qcdl/components.py
```diff
@@ -76,6 +76,21 @@ def default(self, obj: Any) -> Any:
...
+
+    Args:
+        dtype: ``"int"`` or ``"float"``.
```
JoelPasvolsky, line 83:
```suggestion
        dtype: ``"int"`` for a :class:`~dwave.gate.qcdl.registers.Register` or
            ``"float"`` for a
            :class:`~dwave.gate.qcdl.registers.FixedPointRegister`.
```

This is my guess, the intention is to let the user know what each of the dtypes are meant for.

dwave/gate/qcdl/components.py
```diff
@@ -76,6 +76,21 @@ def default(self, obj: Any) -> Any:
...
+        procedure: Procedure that made the allocation. The name it was made
+            under is reported when a later declaration clashes, and the
+            identity tells a re-run of that same procedure apart from a
+            genuine re-declaration.
```
JoelPasvolsky:
```suggestion
        procedure: The allocating procedure. If its name is then reused,
            reruns of the procedure are distinguished from redeclarations.
```

dwave/gate/qcdl/components.py
```diff
@@ -241,6 +256,104 @@ def register_module_used(self, module_name: str | None) -> None:
...
+        name: str,
+        dtype: str,
+        allow_existing: bool = False,
+        initial_value_specified: bool = False,
+    ) -> None:
+        """Record a register allocation, rejecting a silent re-declaration.
+
+        The compiler keeps the *first* allocation of a name, so a second
+        declaration of the same name on the same module is a no-op: its initial
+        value never reaches the qubit. That is almost always a mistake, so it is
```
JoelPasvolsky: This statement need fixing: "its initial value never reaches the qubit"
The paragraph is talking about ``name`` so this is currently saying the value of ``name`` does not reach the qubit. My guess is that the initial value set for the register is not executed?

dwave/gate/qcdl/components.py
```diff
@@ -241,6 +256,104 @@ def register_module_used(self, module_name: str | None) -> None:
...
+    def register_memory_allocation(
+        self,
+        modules: Sequence[QCDLModule],
+        name: str,
+        dtype: str,
+        allow_existing: bool = False,
+        initial_value_specified: bool = False,
+    ) -> None:
+        """Record a register allocation, rejecting a silent re-declaration.
+
```
JoelPasvolsky, line 268:
```suggestion

        This method is mostly intended for use by developers of QCDL; the
        :class:`~dwave.gate.qcdl.registers.Register` and
        :class:`~dwave.gate.qcdl.registers.FixedPointRegister` classes call it
        for you.

```

Moved up so only relevant users read

dwave/gate/qcdl/components.py
```diff
@@ -241,6 +256,104 @@ def register_module_used(self, module_name: str | None) -> None:
...
+        existing memory is wanted, whereas giving a value says the opposite,
+        and the compiler would ignore it. This applies only once the name is
+        allocated; a first allocation always takes its value, whatever the
+        caller opted in to.
+
+        This method is mostly intended for use by developers of QCDL; the
+        :class:`~dwave.gate.qcdl.registers.Register` and
+        :class:`~dwave.gate.qcdl.registers.FixedPointRegister` classes call it
+        for you.
+
```
JoelPasvolsky, line 273:
```suggestion
```

Moved up

dwave/gate/qcdl/components.py
```diff
@@ -241,6 +256,104 @@ def register_module_used(self, module_name: str | None) -> None:
...
+        being built, but is emitted once, so a declaration reached through a
+        later run of the *same* procedure is not a re-declaration and is not
+        reported.
+
+        Re-declaring the name is allowed when the caller asked for it, but only
+        without an initial value: opting in to the re-declaration says the
+        existing memory is wanted, whereas giving a value says the opposite,
+        and the compiler would ignore it. This applies only once the name is
+        allocated; a first allocation always takes its value, whatever the
+        caller opted in to.
```
JoelPasvolsky: I find it hard to follow the above three paragraphs. Would you try a rewrite from the perspective of a user and I could then take a pass?

dwave/gate/qcdl/components.py
```diff
@@ -241,6 +256,104 @@ def register_module_used(self, module_name: str | None) -> None:
...
+
+        This method is mostly intended for use by developers of QCDL; the
+        :class:`~dwave.gate.qcdl.registers.Register` and
+        :class:`~dwave.gate.qcdl.registers.FixedPointRegister` classes call it
+        for you.
+
+        Args:
+            modules: Modules the register is allocated on.
```
JoelPasvolsky, line 296:
```suggestion
            modules: Modules, typically qubits, the register is allocated on.
```

dwave/gate/qcdl/components.py
```diff
@@ -241,6 +256,104 @@ def register_module_used(self, module_name: str | None) -> None:
...
+            name: Name of the register.
+            dtype: ``"int"`` or ``"float"``.
```
JoelPasvolsky, line 298:
```suggestion
        dtype: ``"int"`` for a :class:`~dwave.gate.qcdl.registers.Register` or
            ``"float"`` for a
            :class:`~dwave.gate.qcdl.registers.FixedPointRegister`.
```

This is my guess, the intention is to let the user know what each of the dtypes are meant for.

dwave/gate/qcdl/components.py
```diff
@@ -241,6 +256,104 @@ def register_module_used(self, module_name: str | None) -> None:
...
+            allow_existing: If True, an existing allocation of ``name`` is
+                accepted as long as no initial value was given. Set by the
+                ``alias`` and ``ignore_reallocation`` arguments of a register.
+                It has no effect when ``name`` is not already allocated.
```
JoelPasvolsky, line 302:
```suggestion
            allow_existing: If True, allow a register declaration that reuses
                ``name`` if no initial value is specified. You set this by specifying
                the ``alias`` and ``ignore_reallocation`` arguments in a
                :class:`~dwave.gate.qcdl.registers.Register` or
                :class:`~dwave.gate.qcdl.registers.FixedPointRegister`
                instantiation. Ignored when ``name`` is not already allocated.
```

dwave/gate/qcdl/components.py
```diff
@@ -241,6 +256,104 @@ def register_module_used(self, module_name: str | None) -> None:
...
+                previous = None
+            if previous is not None and not (
+                allow_existing and not initial_value_specified
+            ):
+                if allow_existing:
+                    raise QCDLUserError(
+                        f"register {name!r} is already allocated on"
+                        f" {module.qcdl_module_name} with dtype"
+                        f" {previous.dtype} in procedure"
+                        f" {previous.procedure.name}, and the compiler keeps"
```
JoelPasvolsky: The next few lines seem too much implementation for an error message, can we cut this short and skip to "do not specify an initial value"?

dwave/gate/qcdl/components.py
```diff
@@ -241,6 +256,104 @@ def register_module_used(self, module_name: str | None) -> None:
...
+                    f" global to the circuit and the compiler keeps the first"
+                    f" allocation, so this one would be discarded. Reuse the"
+                    f" existing register, pick another name, or redeclare it"
+                    f" deliberately with alias=True or ignore_reallocation=True"
+                    f" and no initial value."
+                )
+            allocated[name] = RegisterAllocation(dtype, self)
+
+    def _is_rerun_of(self, other: Procedure) -> bool:
+        """Whether ``other`` is an earlier run of the procedure ``self`` is.
```
JoelPasvolsky:
```suggestion
        """Whether ``other`` is an earlier run of this procedure.
```

dwave/gate/qcdl/components.py
```diff
@@ -241,6 +256,104 @@ def register_module_used(self, module_name: str | None) -> None:
...
+
+        Calling a procedure runs its body again, so a register it declares is
+        seen once per call even though the procedure is emitted once. Those
```
JoelPasvolsky: "emitted" --> does this mean "defined", "instantiated", something else?

dwave/gate/qcdl/components.py
```diff
@@ -241,6 +256,104 @@ def register_module_used(self, module_name: str | None) -> None:
...
+        runs are separate :class:`.Procedure` instances sharing a name, and the
+        name is what the rest of the circuit deduplicates on, so matching on it
+        here agrees with what ends up in the program.
```
JoelPasvolsky: name is what the rest of the circuit **deduplicates** on, so matching on it
        here agrees with what ends up in the program.

This sentence is unclear to me

dwave/gate/qcdl/components.py
```diff
@@ -1230,7 +1343,9 @@ def all_to_all_use(q0, q1):
                     sc = Scope(q0, q1)
                     r1 = sc.Register(name="r1")
                     h(q0)
-                    measure(q0, register=q0.Register(name="r1"))
+                    # alias=True reuses the memory r1 already allocated, so the
+                    # outcome is stored on q0 only rather than mirrored
```
JoelPasvolsky: This seems to suggest that setting ``alias=False`` would mirror the outcome to other qubit registers.

dwave/gate/qcdl/qcdl_circuit.py
```diff
@@ -485,6 +490,19 @@ def set_or_check_nondeterministic_modules(
...
+    @property
+    def allocated_registers(self) -> dict[str, dict[str, RegisterAllocation]]:
+        """Register names allocated so far, by module name then register name.
+
+        Register names are global to a circuit: a name allocated in one
+        procedure is the same memory as that name in another, and only the
+        first allocation of it takes effect. The
+        :meth:`~dwave.gate.qcdl.components.Procedure.register_memory_allocation`
+        method maintains this, and reports a second allocation rather than
+        letting it be discarded.
```
JoelPasvolsky:
```suggestion
        discarding it.
```

Can we say something about how that report is used? It seems to me from the code below, that what is actually meant by "reports a second allocation" is in fact "raises an error", yes?

dwave/gate/qcdl/registers.py
```diff
@@ -257,6 +264,24 @@ def _register_initialization(
             name, initial_value=initial_value, length=length, dtype=dtype, signed=signed
         )
 
+        if alias is True and initial_value_specified:
+            raise QCDLUserError(
+                f"register {name!r} is an alias, so no memory is allocated for"
+                f" it and the initial value given here would never reach the"
+                f" qubit; drop the initial value, or drop alias=True to"
+                f" allocate new memory"
```
JoelPasvolsky: A more typical error message just says something like "f"register {name!r} is an alias, so you cannot specify an initial value"

dwave/gate/qcdl/registers.py
```diff
@@ -655,16 +680,24 @@ class Register(IntegerOpsMixin, AssignmentOpsMixin, RegisterInitializerMixin, Ta
             one or more qubits. Typically, you create a register from a
             :class:`~dwave.gate.qcdl.Scope` object, which handles this parameter
             for you.
-        initial_value: Initial value. Defaults to 0.
+        initial_value: Initial value. Defaults to 0. Only the first allocation
+            of a name takes effect, so a value given for a name that is already
```
JoelPasvolsky:
```suggestion
            of a name takes effect, so a value given for a ``name`` that is already
```

dwave/gate/qcdl/registers.py
```diff
@@ -655,16 +680,24 @@ class Register(IntegerOpsMixin, AssignmentOpsMixin, RegisterInitializerMixin, Ta
...
         master_kwargs: Propagate this to the master instruction.  This parameter
             is intended for use by developers of QCDL.
         alias: Set to True if you are aliasing an existing register, and for
             type punning reuse that register's name in the ``name`` parameter.
-            Aliased registers are not reinitialized.
-        ignore_reallocation: If True, the compiler does not reallocate if
-            already allocated (and does not raise an exception).
+            Aliased registers are not reinitialized, so you may not give an
+            ``initial_value``.
+        ignore_reallocation: If True, and the name is already allocated, the
```
JoelPasvolsky:
```suggestion
        ignore_reallocation: If True, and the ``name`` is already allocated, the
```

dwave/gate/qcdl/registers.py
```diff
@@ -655,16 +680,24 @@ class Register(IntegerOpsMixin, AssignmentOpsMixin, RegisterInitializerMixin, Ta
...
+            compiler does not reallocate it (and does not raise an exception).
+            In that case you may not give an ``initial_value``, since it would
+            never reach the qubit. A name that is not yet allocated is
+            allocated as usual and may carry a value.
```
JoelPasvolsky:
```suggestion
```

I think it's better to keep this info in the relevant args

dwave/gate/qcdl/registers.py
```diff
@@ -772,16 +805,24 @@ class FixedPointRegister(
             one or more qubits. Typically, you create a register from a
             :class:`~dwave.gate.qcdl.Scope` object, which handles this parameter
             for you.
-        initial_value: Initial value. Defaults to 0.0.
+        initial_value: Initial value. Defaults to 0.0. Only the first
+            allocation of a name takes effect, so a value given for a name that
```
JoelPasvolsky:
```suggestion
            allocation of a ``name`` takes effect, so a value given for a name that
```

dwave/gate/qcdl/registers.py
```diff
@@ -772,16 +805,24 @@ class FixedPointRegister(
...
+            is already allocated is rejected rather than silently discarded;
+            see ``ignore_reallocation``. An ``alias`` never allocates, so it
+            never takes a value at all.
```
JoelPasvolsky:
```suggestion
            is already allocated is rejected; see ``ignore_reallocation``
```

dwave/gate/qcdl/registers.py
```diff
@@ -772,16 +805,24 @@ class FixedPointRegister(
...
-            Aliased registers are not reinitialized.
-        ignore_reallocation: If True, the compiler does not reallocate if
-            already allocated (and does not raise an exception).
+            Aliased registers are not reinitialized, so you may not give an
+            ``initial_value``.
+        ignore_reallocation: If True, and the name is already allocated, the
+            compiler does not reallocate it (and does not raise an exception).
+            In that case you may not give an ``initial_value``, since it would
+            never reach the qubit. A name that is not yet allocated is
+            allocated as usual and may carry a value.
```
JoelPasvolsky:
```suggestion
```

JoelPasvolsky (approved):

dwave/gate/qcdl/components.py
qci-amos: cbb8f2c

dwave/gate/qcdl/components.py
qci-amos: There's also `Array`... I think it's ok to leave it non-specific as "register"?

dwave/gate/qcdl/components.py
qci-amos: cbb8f2c

dwave/gate/qcdl/components.py
qci-amos: I want to keep it abstract as "modules" here because it could be couplers and we want to hide these details in the abstraction.

dwave/gate/qcdl/components.py
qci-amos: b8b8adc

dwave/gate/qcdl/components.py
qci-amos: 66eb3e5

dwave/gate/qcdl/components.py
qci-amos: also 66eb3e5

qci-amos: I'm going to go ahead and resolve this conversation and defer to https://github.com/dwavesystems/dwave-gate/pull/73/changes#r3823408037
(or new conversations) to continue the discussion.

dwave/gate/qcdl/components.py
qci-amos: It's also used for `Array` (and other contexts like arbitrary function also have a dtype). Are you comfortable just leaving it abstract with "register" here?

dwave/gate/qcdl/components.py
qci-amos: same as https://github.com/dwavesystems/dwave-gate/pull/73/changes#r3854363887

dwave/gate/qcdl/components.py
qci-amos: e8aaa5c

dwave/gate/qcdl/components.py
qci-amos: this is referring to the process of converting the user's Python code to `QCDLProcedureDef`

qci-amos: 04a4f44

dwave/gate/qcdl/components.py
qci-amos: 04a4f44

dwave/gate/qcdl/components.py
qci-amos: resolving this comment to consolidate further discussion to https://github.com/dwavesystems/dwave-gate/pull/73/changes#r3823775764

(or new comments)

dwave/gate/qcdl/components.py
qci-amos: 15e1b97

qci-amos: a31a4f9

dwave/gate/qcdl/qcdl_circuit.py
qci-amos: 27bfabd

dwave/gate/qcdl/components.py
JoelPasvolsky: Sure

dwave/gate/qcdl/components.py
JoelPasvolsky: We have been using the phrase with the "typically" for exactly that purpose for this first release. The idea being to help a new user understand what a module might be (we never replace "module" with just "qubit" in such places).
Up to you

dwave/gate/qcdl/components.py
JoelPasvolsky: Sure

PR #74, opened 2026-08-19
QCDL signals: accept any group of qubits as one_to_all destinations

qci-amos:

### The problem

`QCDLModule.one_to_all` was annotated and written for a `Scope` only, reaching
straight for `destinations.qcdl_modules` and `destinations.scope_id`. Passing
anything else — including `scope.qcdl_modules`, which is what the user guide
shows in places — raised `AttributeError: 'list' object has no attribute
'scope_id'`. Sending to a single qubit meant wrapping it in a `Scope` first.

### The change

`_as_modules` normalizes the destinations into a list of modules, accepting a
`Scope`, a single `QCDLModule`, or a sequence of either, and deduplicating by
name while preserving order.

`scope_id` is only carried through when a `Scope` was actually passed: a bare
sequence has no identity of its own, so tagging its statements with someone
else's scope id would be a lie. That distinction is covered by a test.

A non-module destination, or an empty one, now raises `QCDLUserError` naming
the parameter, rather than surfacing as an `AttributeError` on an internal
attribute further down.

### Compatibility

Purely widening — every call that worked before works unchanged and produces
the same statement.

### Testing

`pytest tests/` passes. `tests/test_scope.py` parameterizes over each accepted
destination form (scope, list, tuple, reversed list, single module, mixed,
repeated) and asserts the resulting qubit list, plus the `scope_id` rule and
the three rejection cases.


dwave/gate/qcdl/components.py
```diff
@@ -1834,7 +1882,10 @@ def signal(self) -> str:
...
 
     def one_to_all(
-        self, destinations: Scope, send: RegisterExpression, **kwargs: Any
+        self,
+        destinations: Scope | QCDLModule | Sequence[QCDLModule],
+        send: RegisterExpression,
+        **kwargs: Any,
     ) -> None:
         """Send a bit from one qubit to a :class:`~dwave.gate.qcdl.Scope` of
         other qubits.
```
JoelPasvolsky:
```suggestion
"""Send a bit from one qubit to other qubits.
```

dwave/gate/qcdl/components.py
```diff
@@ -1843,19 +1894,25 @@ def one_to_all(
...
-            destinations: The scope of all qubits to send the message to. The
-                bit is placed on each qubit's branch condition to be used in a
-                conditional statement.
+            destinations: The qubits to send the message to, as a
+                :class:`~dwave.gate.qcdl.Scope`, a single
+                :class:`~dwave.gate.qcdl.QCDLModule`, or a sequence of either.
+                The bit is placed on each qubit's branch condition to be used in
+                a conditional statement. Statements are tagged with the
+                ``scope_id`` only when a :class:`~dwave.gate.qcdl.Scope` is
+                passed.
```
JoelPasvolsky:
```suggestion
            destinations: The qubits to send the message to. The bit is placed
                on each qubit's branch condition to be used in a conditional
                statement. Statements are tagged with the
                ``scope_id`` if a :class:`~dwave.gate.qcdl.Scope` is
                specified.
```

The typing stands in for the "as a :class:`~dwave.gate.qcdl.Scope`, a single :class:`~dwave.gate.qcdl.QCDLModule`, or a sequence of either"

JoelPasvolsky (approved):

dwave/gate/qcdl/components.py
```diff
@@ -1834,28 +1882,34 @@ def signal(self) -> str:
...
     def one_to_all(
-        self, destinations: Scope, send: RegisterExpression, **kwargs: Any
+        self,
+        destinations: Scope | QCDLModule | Sequence[QCDLModule],
+        send: RegisterExpression,
+        **kwargs: Any,
     ) -> None:
-        """Send a bit from one qubit to a :class:`~dwave.gate.qcdl.Scope` of
-        other qubits.
+"""Send a bit from one qubit to other qubits.
```
qci-amos:
```suggestion
        """Send a bit from one qubit to other qubits.
```

releasenotes/notes/one-to-all-destination-forms-451c4c31fff6e100.yaml
```diff
@@ -0,0 +1,22 @@
...
+fixes:
+  - |
+    Fix ``QCDLModule.one_to_all`` raising an ``AttributeError`` when
+    ``destinations`` is a sequence of qubits, such as the ``qcdl_modules`` of a
+    ``Scope``, rather than the ``Scope`` itself.
+upgrade:
+  - |
+    ``QCDLModule.one_to_all`` tags its statements with a ``scope_id`` only when
+    ``destinations`` is a ``Scope``; a bare sequence of qubits has no identity
+    of its own and so carries no ``scope_id``.
```
thisac: This is a consequence of the fix above, no? Since there was never a case where you could pass a "bare sequence of qubits" to `one_to_all()` before, so either bake this into the above fix or remove.

releasenotes/notes/one-to-all-destination-forms-451c4c31fff6e100.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - |
+    ``QCDLModule.one_to_all`` now raises ``QCDLUserError`` if ``destinations``
+    is not a group of qubits, contains something that is not a qubit, or holds
+    no qubits at all. Its docstring previously advertised ``ValueError``, which
+    it did not raise.
```
thisac: Since we usually don't add smaller doc fixes to release notes, I'd also skip this. It's again a consequence of the feature and fix above.
```suggestion
```

dwave/gate/qcdl/components.py
```diff
@@ -1666,6 +1666,54 @@ def run_prog(idx: int, arg: Any) -> None:
         return shape, table_row
 
 
+def _as_modules(qubits: Any, argument: str) -> tuple[list[QCDLModule], int | None]:
```
thisac:
```suggestion
def _as_modules(qubits: Scope | QCDLModule | Sequence[QCDLModule], argument: str) -> tuple[list[QCDLModule], int | None]:
```

dwave/gate/qcdl/components.py
```diff
@@ -1666,6 +1666,54 @@ def run_prog(idx: int, arg: Any) -> None:
...
+    """Normalize a group of qubits into modules and a scope id.
+
+    Accepts a :class:`.Scope`, a single :class:`.QCDLModule`, or a sequence of
+    either, so that a caller does not have to know which of those an API wants.
+
+    Args:
+        qubits: The group of qubits.
+        argument: Name of the parameter being normalized, for error messages.
```
thisac, line 1681: IMO this isn't really needed. I would seldom care about the name of the variable passed into the function (since that's not easily accessible within a function scope). Alternatively, since this is only used in here and private, maybe just hard-code "destinations", but I'd prefer just skipping this.

thisac (approved):
A few minor comments, but otherwise LGTM.

releasenotes/notes/one-to-all-destination-forms-451c4c31fff6e100.yaml
qci-amos:
```suggestion
```

dwave/gate/qcdl/components.py
qci-amos: 62bffb0

dwave/gate/qcdl/components.py
qci-amos: 802e45a

PR #75, opened 2026-08-19
QCDL: validate the arguments and signature of the qcdl decorator

qci-amos:

### The problem

The `qcdl` decorator injects qubits into the decorated function **by name**, as
keyword arguments `q0`, `q1`, ... Three things went wrong quietly:

- A parameter that is not a `q<N>` receives nothing. Python's own message —
  `main() missing 1 required positional argument: 'alpha'` — gives no hint that
  qubits are involved or that the name is the reason.
- A generated qubit with no parameter to land in was dropped from the program
  without a word: `@qcdl(3)` on `def main(q0, q1)` built a two-qubit program.
- `num_qubits` was never checked. `@qcdl(0)`, `@qcdl(2.5)` and a bare `@qcdl`
  (which passes the decorated function in as the qubit count) all failed later,
  obscurely or not at all.

### The change

`_validate_num_qubits` checks the count up front, including the bare-`@qcdl`
case, which gets its own message telling you to call the decorator.

At call time, any required parameter the call would leave unbound is reported
with a `TypeError` that keeps Python's own wording and then explains the rule:
which qubits were supplied, that they are injected as `q<N>`, and whether the
fix is to raise `num_qubits` or to rename the parameter. Separately, a
generated qubit with no parameter for it raises `QCDLUserError` rather than
being dropped.

Two cases are deliberately left alone:

- a signature taking `**kwargs` absorbs every qubit, so nothing is dropped;
- an environment supplies its whole qubit set by definition, so a signature
  naming fewer is expected and still allowed.

### Compatibility

Programs that were silently losing a qubit, or that would have raised a
confusing `TypeError` anyway, now fail with an explanation. Everything that
built the program the author intended is unchanged.

### Testing

`pytest tests/` passes. `tests/test_qcdl_circuit.py` gains two classes covering
`num_qubits` validation and each signature shape: dropped qubit, `**kwargs`,
environment mode, non-`q<N>` parameter, extra qubit parameter, several unfilled
parameters at once, keyword-only parameters, caller-supplied values, and
inferred mode with a sparse set (`q0`, `q5`).


dwave/gate/qcdl/qcdl_circuit.py
```diff
@@ -551,6 +552,116 @@ def _get_fspec(f: Any) -> tuple[list[str], str | None]:
...
+        :exception:`~dwave.gate.qcdl.exceptions.QCDLUserError`: If
+            ``num_qubits`` could not generate at least one qubit.
+    """
+    if callable(num_qubits):
+        raise QCDLUserError(
+            f"the qcdl decorator must be called, so decorate"
+            f" {getattr(num_qubits, '__name__', num_qubits)} with @qcdl() or"
+            f" @qcdl(num_qubits) rather than with a bare @qcdl"
+        )
+    if isinstance(num_qubits, bool) or not isinstance(num_qubits, numbers.Integral):
```
JoelPasvolsky, line 568: Curious what the reason is to not do: ``type(num_qubits) is not int``

dwave/gate/qcdl/qcdl_circuit.py
```diff
@@ -551,6 +552,116 @@ def _get_fspec(f: Any) -> tuple[list[str], str | None]:
...
+    num_qubits: int | None,
+    from_environment: bool,
+) -> TypeError:
+    """Build the error for an entry point whose parameters were not all filled.
+
+    The decorator injects qubits by *name*, so a mistyped or differently named
+    parameter silently receives nothing. Python's own message for that names
+    only the parameter, which gives no hint that qubits are involved.
+    """
+    plural = "" if len(missing) == 1 else "s"
```
JoelPasvolsky: This seems a big code investment compared to ``required positional argument(s)`` for line 631. Is it worth complicating the code for such a thing?

dwave/gate/qcdl/qcdl_circuit.py
```diff
@@ -551,6 +552,116 @@ def _get_fspec(f: Any) -> tuple[list[str], str | None]:
...
+    supplied_names = sorted(supplied, key=lambda name: (len(name), name))
+    message += (
+        f" The qcdl decorator injects qubits as keyword arguments named q<N>"
+        f" (q0, q1, ...), and {source} supplied"
+        f" {', '.join(supplied_names) if supplied_names else 'none'}."
+    )
+
+    if any(is_qubit_or_coupler_name(name) for name in missing):
+        message += (
+            " Raise num_qubits to cover the missing qubits, or drop the"
```
JoelPasvolsky: Just checking: is this correct? Haven't followed it carefully down the stack but if sufficient qubits are not named should the qubit number not be lowered?

dwave/gate/qcdl/qcdl_circuit.py
```diff
@@ -551,6 +552,116 @@ def _get_fspec(f: Any) -> tuple[list[str], str | None]:
...
+            " parameters for them."
+        )
+    else:
+        matches = "does not match" if len(missing) == 1 else "do not match"
```
JoelPasvolsky: Same comment: for simpler code ``" not a match for that pattern"`` seems  good enough for an eror message. Also, for the ``plural`` used here and taken from above

dwave/gate/qcdl/qcdl_circuit.py
```diff
@@ -732,6 +848,39 @@ def wrapper(*args: Any, **kwargs: Any) -> QCDLProgram | QCDLV2:
...
+
+            if num_qubits is not None and not _env and not f_keywords:
+                dropped = [q for q in module_kwargs if q not in f_args]
+                if dropped:
+                    raise QCDLUserError(
+                        f"@qcdl({num_qubits}) generates"
+                        f" {', '.join(module_kwargs)} but"
+                        f" {getattr(f, '__name__', 'the decorated function')}()"
+                        f" has no parameter for {', '.join(dropped)}, so"
+                        f" {'they' if len(dropped) > 1 else 'it'} would be"
```
JoelPasvolsky:
```suggestion
                        f" would be"
```

If accepted might be easier to do manually to keep a good line length

JoelPasvolsky (approved):

dwave/gate/qcdl/qcdl_circuit.py
qci-amos: My viewpoint on this sort of consideration has changed dramatically the last several months. Last year I would have agreed with you with a passion, but now... if ai is "willing" to do the grunt work, and back it up with tests, then why not just get the best error message possible? Furthermore, in my experience occasionally subtle changes in wording can help someone providing support narrow down what went wrong.

arcondello: Ultimately this is up to the maintainers of this repo to decide. But my unsolicited 2c: this is the kind of Claude-ism that I would tell it not do to.

Reasons I don't like it:
- I don't think we've left the era of code being read more than written. And even if it's not by much, this does add additional cognitive load to the reader, whether that reader is human or an LLM.
- It's not infrequent that when an error message is raised, I search the exact string of the error message in the code. There are times that the traceback is not available for whatever reason. And Claude does the same thing. I don't love needing to move to a fuzzy search.
- I suppose if I wanted to `try`/`catch` this exact error having it change based on a parameter is bad. I will admit this reason is quite weak

Of course all of that needs to be weighed against the potential benefit to the user. IMO in this case `(s)` is equally readable but :shrug:

dwave/gate/qcdl/qcdl_circuit.py
qci-amos: this exception covers the case of:

```python
@qcdl(num_qubits=1)
def main(q0, q1)
```

JoelPasvolsky: Makes sense, thanks!

dwave/gate/qcdl/qcdl_circuit.py
qci-amos: `numbers.Integral` covers more cases, e.g., numpy ints but `isinstance(True, int)` is `True`

dwave/gate/qcdl/qcdl_circuit.py
qci-amos: I asked Claude to simplify the validation. I asked it to try pydantic but it explained why that wouldn't work in this context. But it found other ways to make this simpler and it looks like the error message in question ended up getting simplified too:
827acee90035f879a81171bf1a578c969ec63158

in any case, the current wording of this error message is I believe warranted due to the complexity of using a decorator and the conventions we've been using for arguments passed to the decorated function.

dwave/gate/qcdl/qcdl_circuit.py
qci-amos: 998aa7cb1c5732b1f830e3ecd1017e335eab4ede

dwave/gate/qcdl/qcdl_circuit.py
qci-amos: this was refactored

dwave/gate/qcdl/qcdl_circuit.py
```diff
@@ -551,6 +552,96 @@ def _get_fspec(f: Any) -> tuple[list[str], str | None]:
...
 
 
+def _validate_num_qubits(num_qubits: Any) -> None:
+    """Check the ``num_qubits`` argument of the :func:`qcdl` decorator.
+
+    Raises:
+        :exception:`~dwave.gate.qcdl.exceptions.QCDLUserError`: If
+            ``num_qubits`` could not generate at least one qubit.
+    """
+    if callable(num_qubits):
```
thisac, line 562: It took me a second to understand how this test was done. Might be good to add a comment about `num_qubits` being the wrapped function (and thus callable) instead of the actual number of qubits integer.

thisac (approved):
Seems like a very reasonable validation update to make.

dwave/gate/qcdl/circuit.py
```diff
@@ -551,6 +552,96 @@ def _get_fspec(f: Any) -> tuple[list[str], str | None]:
...
 
 
+def _validate_num_qubits(num_qubits: Any) -> None:
+    """Check the ``num_qubits`` argument of the :func:`qcdl` decorator.
+
+    Raises:
+        :exception:`~dwave.gate.qcdl.exceptions.QCDLUserError`: If
+            ``num_qubits`` could not generate at least one qubit.
+    """
+    if callable(num_qubits):
```
qci-amos, line 558: a6d50e4

PR #76, opened 2026-08-19
results: register defaults, YieldHandling key formats, and root exports

qci-amos:
Part of splitting #71 into reviewable pieces. This is the non-QCDL half —
`dwave.gate.results` — and is independent of the other four PRs.

Three independent fixes, one commit each.

### 1. `get_memory` / `get_counts` default to the measured register

Both fell back to `get_default_register`, which lists **every** qubit in the
circuit whether or not it was measured. An unmeasured qubit was therefore
padded into the bitstring with `unmeasured_value` and read as an outcome:
a three-qubit circuit measuring only `q1` and `q2` returned `{"01_": 3}`.

They now default to `get_measurements_register`, falling back to every qubit
only when nothing was measured at all. `get_default_register` is unchanged and
still available — pass it explicitly to get the padded register back — and its
docstring now says which methods use it and when.

**Compatibility:** this changes the default output shape for any circuit with
unmeasured qubits. Callers passing an explicit `register` are unaffected.

### 2. `YieldHandling.apply` handles every counts key format

`apply` tested keys with `"*" not in k`. `count_measurements(..., key_format=None)`
produces **integer** keys, so composing the two — which is the obvious thing to
do — raised `TypeError: argument of type 'int' is not iterable`. The
all-erasures fallback had the same assumption, building its zeros key with
`str.replace`.

Only the string key formats can hold a splat, so the erasure test now checks
for that, and the fallback key falls back to `0` for a non-string key.

An empty distribution raised `StopIteration` out of that fallback, which is not
a useful error for a counts dict that post-selection emptied; it now raises
`ValueError`.

`apply` also picked up the docstring it never had.

### 3. Export `Result` and `YieldHandling` from the package root

The user guide refers to these as `dwave.gate.Result` and
`dwave.gate.YieldHandling`, which did not resolve. They are re-exported from
the package root and remain importable from `dwave.gate.results`, which is
still where they are defined — a test asserts `__module__` does not move, since
Sphinx and pickle both key off it.

### Testing

`pytest tests/` passes. `tests/test_results.py` covers the register default
(including the explicit-register and nothing-measured paths), `apply` against
every key format `count_measurements` produces, and the import paths.


dwave/gate/results.py
```diff
@@ -638,16 +652,53 @@ def from_name(name: str | YieldHandling) -> YieldHandling:
         else:
             return YieldHandling[name]
 
+    @staticmethod
+    def _is_erasure(key: Any) -> bool:
+        """Whether a counts key records an erasure.
+
+        Only the string key formats can hold a splat: the
+        :func:`.count_measurements` function produces integer keys only for
+        memory that is entirely 0s and 1s.
```
JoelPasvolsky, line 661: wouldn't splay be held as -1? Maybe not here.

dwave/gate/results.py
```diff
@@ -638,16 +652,53 @@ def from_name(name: str | YieldHandling) -> YieldHandling:
...
+            distribution: Counts, as returned by the
+                :func:`.count_measurements` function or the
+                :meth:`.Result.get_counts` method. Keys may be in any of the
+                formats that function produces.
+
+        Raises:
+            :exception:`ValueError`: If the distribution is empty, or if the
+                yield is too low for
+                :attr:`.renormalize_distribution_or_raise`.
+            :exception:`ZeroDivisionError`: If nothing survives post selection
```
JoelPasvolsky, line 680: Is this the best error type for this case?

dwave/gate/results.py
```diff
@@ -638,16 +652,53 @@ def from_name(name: str | YieldHandling) -> YieldHandling:
...
     def apply(
         self, distribution: dict[Any, int]
     ) -> tuple[dict[Any, float | int], float]:
+        """Apply this yield handling to a counts dict.
+
+        Args:
+            distribution: Counts, as returned by the
+                :func:`.count_measurements` function or the
+                :meth:`.Result.get_counts` method. Keys may be in any of the
+                formats that function produces.
```
JoelPasvolsky, line 674: by "formats" do you mean 0, 1, and *?

dwave/gate/results.py
```diff
@@ -638,16 +652,53 @@ def from_name(name: str | YieldHandling) -> YieldHandling:
...
+
+        Raises:
+            :exception:`ValueError`: If the distribution is empty, or if the
+                yield is too low for
+                :attr:`.renormalize_distribution_or_raise`.
+            :exception:`ZeroDivisionError`: If nothing survives post selection
+                and the distribution has to be renormalized.
+
+        Returns:
+            The handled distribution and the observed yield.
```
JoelPasvolsky, line 684:
```suggestion
            The renormalized distribution and observed yield.
```

Based on ``return renormalized, observed_yield``

JoelPasvolsky (approved):

dwave/gate/results.py
qci-amos: What's at stake here are the data types that `count_measurements` can return. The `key_format` there may be `bin`, `hex`, and `None`. Maybe we should remove `hex` and `None`, I don't know of any current use of those (`count_measurements` was written before erasures).

But regardless, `hex` and "integer" for keys don't support splats. We only have `-1` outside of a bitstring context or `*` for a bitstring context.

dwave/gate/results.py
qci-amos: It's raised here:
https://github.com/dwavesystems/dwave-gate/blob/5d1f965b2c53abceae45dc699c7f0983b655973c/dwave/gate/results.py#L711

I generally try to use the built-in types where it makes sense and it seems like this qualifies: there really is a "divide by zero" that would occur here. We could do a custom error if that's what you mean?

dwave/gate/results.py
qci-amos: this is the same context as this comment:
https://github.com/dwavesystems/dwave-gate/pull/76#discussion_r3853944907

dwave/gate/results.py
qci-amos: That said, I can imagine that at scale we'll want some representation more compact than strings... but we're not at the point where that matters yet and Qiskit requires strings afaik.

dwave/gate/results.py
qci-amos: That was just for one of the `return` lines. The other return lines aren't "renormalized" distributions.

dwave/gate/results.py
JoelPasvolsky: I think I meant that our code should prevent an known divide-by-zero error if it's caused by a bad user input (not allow such an input) or a bad calculation on our part (return some message that there are insufficient samples or some such). 
If it means a big change to the application, I don't have a preference for a custom error. See what makes sense to you

qci-amos: The cause of this could be a circuit that's too deep or wide. So it's not that the input is "bad" in a validation sense nor a bug in the qpu/simulator.

We've debated in the past whether this case should be a log message (that might be missed) or a raised exception and we've gone with the later because otherwise downstream logic (e.g., qiskit) might raise their own less-informative exception if they receive an empty bitstring distribution.

dwave/gate/__init__.py
```diff
@@ -13,3 +13,13 @@
 #    limitations under the License.
 
 __version__ = "0.4.0"
+
+# The user guide refers to these as dwave.gate.Result and
+# dwave.gate.YieldHandling, so re-export them from the package root. They remain
+# importable from dwave.gate.results, which is where they are defined.
```
thisac, line 19: This comment doesn't really add any value.
```suggestion
```

dwave/gate/results.py
```diff
@@ -495,6 +503,11 @@ def get_memory(
             tag = self.default_tag
         if self.measurements is None:
             raise RuntimeError("measurements are not available for this result")
+        if register is None:
+            # A qubit with no measurements has no bit to contribute, and padding
+            # one in makes the bitstring look like an outcome. Fall back to every
+            # qubit only when nothing was measured at all.
+            register = cast(RegisterType, self.get_measurements_register(tag)) or None
```
thisac, line 510: It might be correct to use here, but I really don't like `typing.cast`.

thisac: Is it even needed here? `RegisterType` is just `list[str | None]` and `self.get_measurements_register()` returns `list[str]`.

tests/test_results.py
```diff
@@ -36,6 +36,26 @@
...
+    assert dwave.gate.YieldHandling is YieldHandling
+    assert set(dwave.gate.__all__) == {"Result", "YieldHandling"}
+
+
+def test_re_export_leaves_the_defining_module_alone():
+    """Sphinx and pickle both key off __module__, so it must not move."""
+    assert Result.__module__ == "dwave.gate.results"
+    assert YieldHandling.__module__ == "dwave.gate.results"
+
+
```
thisac, line 58: I would probably skip these tests since we're not usually testing every single import or namespace. Seems a bit excessive.

qci-amos: 4018190

dwave/gate/results.py
qci-amos: AI thinks it's necessary because the underlying type is (mutable) `list`. I think we should switch that to `tuple` (but allow users to continue to pass `list`). I asked AI to try to switch to `tuple` but it's not doing it quite the way I want... if you'd like I could add that to this PR otherwise I propose we defer.

releasenotes/notes/results-api-defaults-and-key-handling-4d2fdb5d28df4920.yaml
```diff
@@ -0,0 +1,31 @@
+---
+features:
+  - |
+    Re-export ``Result`` and ``YieldHandling`` from the package root, so that
+    ``dwave.gate.Result`` and ``dwave.gate.YieldHandling``, as the user guide
+    refers to them, now resolve. Both remain importable from
+    ``dwave.gate.results``, which is still where they are defined.
```
thisac: Not addressed here.
```suggestion
```

thisac (approved):
LGTM, apart from a release-notes entry that should be removed.

PR #77, opened 2026-08-21
Add `LeapQCDLSimulator`, an Ocean sampler-like interface for the QCDL solver

randomir:
In this PR we implement the higher-level, sampler-like interface for the QCDL solver in Leap.

The class name, `LeapQCDLSimulator` is up for discussion. @qci-amos initially suggested `QCDLSimulator`, I added `Leap` for consistency with other Ocean samplers in Leap.

We use `run()` method instead of `sample()` for consistency with other gate-model libraries.

~One thing I'll probably change is the result type contained in the `run()`-returned future. Currently, that's just `dwave.gate.results.Result`, but for convenience we should add some metadata, like the timing info returned, the job id, etc. We could model the `SampleResult` returned by the Stride sampler.~
(Edit: done in b948ce14dfb217bda896844524b2eae28d97fc1a.)

In addition, we could enrich the returned future with progress-tracking attributes/methods (e.g. upload in progress, upload completed, problem submitted, pending, done, etc).

I'll leave the latter for a follow-up PR.

#### Usage Example

```python
from dwave.gate.qcdl.leap import LeapQCDLSimulator

simulator = LeapQCDLSimulator(profile='test')
future = simulator.run(qcdl_program, shots=100)
result = future.result()

measurements = result.measurements
```

#### AI Generation Disclosure
Used Claude Fable 5 for the first draft of tests (translation of NLSampler tests to the QCDL use case).

### DIFF
--- a/dwave/gate/qcdl/leap.py
+++ b/dwave/gate/qcdl/leap.py
@@ -19,7 +19,7 @@ from __future__ import annotations
 from collections.abc import Mapping
 from concurrent.futures import Future, ThreadPoolExecutor
 from functools import cached_property
-from typing import Any, Self
+from typing import Any, NamedTuple, Self
 
 import orjson
 from dwave.cloud.client import Client
@@ -58,7 +58,7 @@ class LeapQCDLSimulator:
 
             simulator = LeapQCDLSimulator()
             future = simulator.run(qcdl_program, shots=100)
-            result = future.result()
+            result = future.result().result
 
         Access measurements and sample counts by calling
         ``result.measurements`` or ``result.get_counts()`` respectively.
@@ -163,10 +163,18 @@ class LeapQCDLSimulator:
         # raise exceptions from the runtime context as they're not handled here
         return None
 
-    def run(self, qcdl: QCDLProgram | Mapping[str, Any], **params) -> Future[Result]:
+    class RunResult(NamedTuple):
+        """:meth:`~dwave.gate.qcdl.leap.LeapQCDLSimulator.run` method future result
+        """
+        result: Result
+        # note: use a string annotation to avoid dependency on dwave-system
+        info: 'dwave.system.samplers.ResultInfoDict'
+
+    def run(self, qcdl: QCDLProgram | Mapping[str, Any], **params) -> Future[RunResult]:
         """Run the :term:`QCDL` program using the selected Leap simulator,
-        and return the :class:`~dwave.gate.results.Result` in a
-        class:`~concurrent.futures.Future`.
+        and return the :class:`~dwave.gate.results.Result`, alongside the SAPI
+        job metadata, both wrapped in a :class:`~dwave.gate.qcdl.leap.LeapQCDLSimulator.RunResult`
+        and returned in a class:`~concurrent.futures.Future`.
 
         Args:
             qcdl:
@@ -181,9 +189,18 @@ class LeapQCDLSimulator:
         response = self.solver.sample_qcdl(qcdl, **params)
 
         def decode():
+            # decode/construct the QCDL result
             answer = orjson.loads(response.answer_data.read())
-            return Result(**answer)
+            result = Result(**answer)
 
-        result = self._executor.submit(decode)
+            # collect metadata of interest
+            info = dict(
+                timing=response.timing.copy(),
+                problem_id=response.id,
+                problem_label=response.label,
+                problem_data_id=response.data_id,
+            )
 
-        return result
+            return LeapQCDLSimulator.RunResult(result, info)
+
+        return self._executor.submit(decode)
--- a/dwave/gate/qcdl/leap.py
+++ b/dwave/gate/qcdl/leap.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 from collections.abc import Mapping
 from concurrent.futures import Future, ThreadPoolExecutor
+from functools import cached_property
 from typing import Any, Self
 
 import orjson
@@ -66,25 +67,21 @@ class LeapQCDLSimulator:
 
     @property
     def default_solver(self) -> dict[str, str]:
-        """dict: Features used to select the latest accessible QCDL software solver."""
+        """Features used to select the latest accessible QCDL software solver."""
         return dict(supported_problem_types__contains='qcdl',
                     category='software-gate',
                     order_by='-properties.version')
 
-    @property
+    @cached_property
     def properties(self) -> dict[str, Any]:
         """Solver properties as returned by a :term:`SAPI` query.
 
         Solver properties are dependent on the selected solver and subject to
         change; for example, new features may add properties.
         """
-        try:
-            return self._properties
-        except AttributeError:
-            self._properties = properties = self.solver.properties.copy()
-            return properties
+        return self.solver.properties.copy()
 
-    @property
+    @cached_property
     def parameters(self) -> dict[str, list[str]]:
         """Supported parameters.
 
@@ -94,14 +91,10 @@ class LeapQCDLSimulator:
         Solver parameters are dependent on the selected solver and subject to
         change; for example, new features may add parameters.
         """
-        try:
-            return self._parameters
-        except AttributeError:
-            parameters = {param: ['parameters']
-                          for param in self.properties['parameters']}
-            parameters.update(label=[])
-            self._parameters = parameters
-            return parameters
+        parameters = {param: ['parameters']
+                      for param in self.properties['parameters']}
+        parameters.update(label=[])
+        return parameters
 
     def __init__(self, **config):
         r"""Initialize the simulator client instance.
--- a/dwave/gate/qcdl/leap.py
+++ b/dwave/gate/qcdl/leap.py
@@ -1,0 +1,103 @@
+# Copyright 2026 D-Wave
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+"""Provide access to QCDL solvers in Leap."""
+
+from __future__ import annotations
+
+from collections.abc import Mapping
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -40,11 +42,13 @@ Repository = "https://github.com/dwavesystems/dwave-gate.git"
...
-    "numpy==2.4.6",  # latest that support Python 3.11 as of July 2026
-    "pydantic==2.13.4",  # latest as of July 2026
     "addict==2.4.0",
-    "reno==4.1.0",
+    "dwave-cloud-client==0.14.7",
+    "numpy==2.4.6",     # latest that support Python 3.11 as of July 2026
+    "orjson==3.12",     # latest that supports 3.10--3.15 as of August 2026
     "polars==1.43.1",
+    "pydantic==2.13.4", # latest as of July 2026
+    "reno==4.1.0",
```
qci-amos, line 51: Explicit pins means we're not testing the same versions that users are installing. How is this drift managed?

tests/test_leap.py
```diff
@@ -0,0 +1,148 @@
...
+            assert answer.num_qubits == mock_answer['num_qubits']
+            assert answer.simulated_qcdl == mock_answer['simulated_qcdl']
+        finally:
+            simulator.close()
+
+    @unittest.mock.patch('dwave.gate.qcdl.leap.Client', mock_client_factory)
+    def test_close(self):
+        simulator = LeapQCDLSimulator()
+        with unittest.mock.patch.object(
+                simulator.client, 'close', wraps=simulator.client.close) as client_close:
```
qci-amos, line 142: How should we handle black/ruff formatting?

dwave/gate/qcdl/leap.py
```diff
@@ -0,0 +1,196 @@
...
+            raise ValueError("selected solver does not support the 'qcdl' problem type.")
+
+        self._executor = ThreadPoolExecutor()
+
+    def close(self) -> None:
+        """Close the underlying cloud client to release system resources such as
+        threads.
+
+        The method blocks for all the currently scheduled work (sampling
+        requests) to finish.
```
qci-amos, line 146: Normally I'd expect a user to make just one instance and hold it for their entire session -- am I thinking about this the right way? What happens if a user neglects to close?

Part of the reason this is relevant, and something I've puzzled over wrt executors, is that I've found that there are contexts where there isn't an obvious place to call close / use a context manager, e.g., within a qiskit backend.

tests/test_leap.py
```diff
@@ -0,0 +1,148 @@
...
+            assert properties.get('category') == 'software-gate'
+            # the latest of the two matching solvers is preferred
+            assert properties.get('version') == '0.2'
+        finally:
+            simulator.close()
+
+    @unittest.mock.patch('dwave.gate.qcdl.leap.Client', mock_client_factory)
+    def test_solver_selection_override(self):
+        # explicit solver selection takes precedence over the defaults
+        simulator = LeapQCDLSimulator(solver=dict(name='qcdl_sim_v1'))
```
qci-amos, line 72: What would these correspond to? Different simulator images available in leap? For example, if we make a breaking change, this would allow someone to select a prior version?

How are the names determined? e.g., if I know the image name in Leap?

pyproject.toml
thisac: I would argue that these _should_ be pinned to provide known stable environments, as opposed to the `dependencies` above, which should be as permissive as possible (which are here upper-bounded by major version). These pinned versions get updated quite often as part of other PR or maintenance updates.

What we sometimes do, and could do here, is also add tests for latest + earliest supported versions in CI, but it's IMO not always necessary or worth it.

dwave/gate/qcdl/leap.py
```diff
@@ -0,0 +1,196 @@
...
+            result = future.result()
+
+        Access measurements and sample counts by calling
+        ``result.measurements`` or ``result.get_counts()`` respectively.
+
+    """
+
+    @property
+    def default_solver(self) -> dict[str, str]:
+        """dict: Features used to select the latest accessible QCDL software solver."""
```
thisac:
```suggestion
        """Features used to select the latest accessible QCDL software solver."""
```

dwave/gate/qcdl/leap.py
```diff
@@ -0,0 +1,196 @@
...
+        """Solver properties as returned by a :term:`SAPI` query.
+
+        Solver properties are dependent on the selected solver and subject to
+        change; for example, new features may add properties.
+        """
+        try:
+            return self._properties
+        except AttributeError:
+            self._properties = properties = self.solver.properties.copy()
+            return properties
```
thisac: Minor, but a `cached_property` returning `self.solver.properties.copy()` perhaps looks slightly cleaner.

thisac (commented):
`LeapQCDLSimulator` sounds good to me.

pyproject.toml
qci-amos: On one hand, I don't think I've ever had a challenge reproducing locally an issue first discovered in CI and it will be even more easy now with AI. I've had that experience several times now, I don't even carefully look at the failing build, I just copy/paste the error message into AI and it can figure out what went wrong.

On the other hand, my view is we want the most sensitive tests possible/practical so that we find any issues before users do (but also, avoid these new supply chain attacks...). Fully combinatorial tests are clearly not an option, but we can still do some range checking. We've found it necessary to systematically check all the latest versions twice a day, at least for 1st party dependencies. 

That said, we have been adapting our CI workflows to use `uv` and I think `uv.lock` files are a more effective way to manage dependencies in a context like this than groups in `pyproject.toml`. For example, `uv.lock` may specify a version that's known to work, but then in CI we can deliberately test the lower bounds:
https://docs.astral.sh/uv/concepts/resolution/#lower-bounds

We would do something like this:
```yaml
name: CI

on: [push, pull_request]

jobs:
  test:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        resolution: [lowest-direct, highest]
    steps:
      - uses: actions/checkout@v4

      - name: Install uv
        uses: astral-sh/setup-uv@v3

      - name: Install dependencies (${{ matrix.resolution }})
        run: uv sync --resolution ${{ matrix.resolution }}

      - name: Run tests
        run: uv run --resolution ${{ matrix.resolution }} pytest
```




More discussion is needed. I'm sure the Ocean client a lot more stable and we may not need to be as aggressive, or at least, I'm comfortable assuming this for starters.

randomir: We are not very terribly consistent with sweeping strategies across Ocean packages. There are a few reasons for this, and yes, we should discuss them.

However, generally, we tend to either pin the oldest supported versions in requirements.txt/pyproject.toml, or install from package requirements (meaning latest at the time of CI), and then sweep only those dependencies we consider worthy doing so. For example we sweep dwave-cloud-client in dwave-system, we sweep dimod in dwave-cloud-client, etc. That's in addition to sweeping over supported Python versions.

One could argue that by installing latest (from package deps), and sweeping over Python versions, we effectively sweep over a range of package versions (as they differ per Python version). But that's more of a side-effect than intent.

In this PR, I didn't want to make changes to the existing CI, I just added the newly required dependencies.

My recommendation was/is we try to test with the oldest supported -- i.e. lower bound of the "package install requirement", and then also with the latest available; swept over Python versions. One argument against this was that it's finicky/painful to determine/pin lower bounds for packages, especially in combination with other packages, over Python versions.. So testing requirements soon become an unmaintainable mess (just check dwave-hybrid ones). But LLMs could help us with that.

tests/test_leap.py
randomir: And sacrifice our artistic integrity? :smile:

dwave/gate/qcdl/leap.py
randomir: Our Ocean samplers are "scoped resources", see `dimod.Scoped`, an RAII attempt in python. They're heavy, bring in background thread pools, so should be reused if possible. But if not, at least they should be closed to prevent resource leaks.

tests/test_leap.py
randomir: Names are solver names, as visible in Leap dashboard, or returned by SAPI (try `dwave solvers -l`).

Versions might not be relevant here/now, but it's not impossible to have multiple solvers (think QCDL simulators) online, each with a different version. This just tests that `LeapQCDLSimulator` by default selects the latest. This "default" is trivial to override, so nothing here is set in stone.

pyproject.toml
qci-amos: > In this PR, I didn't want to make changes to the existing CI,

I'm good with this for now.

qci-amos (approved):

dwave/gate/qcdl/leap.py
```diff
@@ -0,0 +1,189 @@
...
+            arguments overrides this specification.
+
+        """
+        # default to short-lived session to prevent resets on slow uploads
+        config.setdefault('connection_close', True)
+
+        # prefer the latest QCDL solver available, but allow for an easy
+        # override on any config level above the defaults (file/env/kwarg)
+        defaults = config.setdefault('defaults', {})
+        if not isinstance(defaults, Mapping):
```
arcondello, line 126: Is this an error a user can get to? Or would this only happen if `setdefault()` had a bug?

arcondello (approved):
LGTM

dwave/gate/qcdl/leap.py
randomir: This is to guard from a user passing in a non-dict for `defaults` kwarg, e.g. `LeapQCDLSimulator(defaults=1)`.

dwave/gate/qcdl/leap.py
```diff
@@ -0,0 +1,207 @@
...
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+"""Provide access to QCDL solvers in Leap."""
+
+from __future__ import annotations
+
+from collections.abc import Mapping
+from concurrent.futures import Future, ThreadPoolExecutor
+from dataclasses import dataclass
```
thisac: Not used.
```suggestion
```

thisac (approved):
LGTM. Would this perhaps belong one step up, next to `results.py`, and not in the QCDL module? It's not exactly part of the QCDL implementation.

randomir:
> LGTM. Would this perhaps belong one step up, next to `results.py`, and not in the QCDL module? It's not exactly part of the QCDL implementation.

Perhaps. I was surprised to see `results.py` outside of the `qcdl` module.. and at the same time a bunch of non-qcdl specific stuff under the qcdl module (e.g. things in `qcdl.utils` and even `blacken_qcdl_str`).

I would probably put both results and leap modules under qcdl :shrug:

PR #80, opened 2026-09-04
Add job submission and solver properties & parameters

JoelPasvolsky:
I will update the existing QCDL for `dwave-gate` job submission and result handling in a second PR, creating this one now to give reviewers time. 

@jstubbs-dwave

### DIFF
--- a/docs/workflow.rst
+++ b/docs/workflow.rst
@@ -1304,8 +1304,10 @@ qpu
 
 The QPU to simulate, formatted as a string.
 
-Specified value must be one of the values of the
-:ref:`property_drsim_supported_qpu_strings` property.
+Specified value must be one of the following values:
+
+*   ``simulator``: Ideal simulator.
+*   ``DR17``: Dual-rail QPU with 17 qubits.
 
 Default is to simulate the QPU specified by the
 :ref:`property_drsim_default_qpu` property.
@@ -1421,6 +1423,212 @@ This example applies a noise model.
 
 Simulator Properties
 --------------------
+
+.. _property_drsim_category:
+
+category
+~~~~~~~~
+
+Type of solver, as a string.
+
+*   ``software-gate``: Gate-model simulator.
+
+Example
+-------
+
+>>> from dwave.system import DWaveSampler
+...
+>>> sampler = DWaveSampler()
+>>> sampler.properties["default_readout_thermalization"]   # doctest: +SKIP
+0.0
+
+.. _property_drsim_default_noise_model:
+
+default_noise_model
+~~~~~~~~~~~~~~~~~~~
+
+Default setting for the application of a noise model, as a Boolean.
+
+*   ``True``: A noise model is applied.
+*   ``False``: Simulates an ideal QPU.
+
+Example
+-------
+
+>>> from dwave.system import DWaveSampler
+...
+>>> sampler = DWaveSampler()
+>>> sampler.properties["default_readout_thermalization"]   # doctest: +SKIP
+0.0
+
+.. _property_drsim_default_qpu:
+
+default_qpu
+~~~~~~~~~~~
+
+Default selection of the QPU to simulate, as a string.
+
+*   ``simulator``: Ideal simulator.
+*   ``DR17``: Dual-rail QPU with 17 qubits.
+
+Example
+-------
+
+>>> from dwave.system import DWaveSampler
+...
+>>> sampler = DWaveSampler()
+>>> sampler.properties["default_readout_thermalization"]   # doctest: +SKIP
+0.0
+
+.. _property_drsim_default_repeat_until_shots_requested:
+
+default_repeat_until_shots_requested
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Default setting, as a Boolean, for rerunning the circuit until the requested
+number of measurements is accumulated.
+
+*   ``True``: Repeatedly rerun the circuit.
+*   ``False``: Run the circuit the number of times set by the
+    :ref:`parameter_drsim_shots` parameter.
+
+Example
+-------
+
+>>> from dwave.system import DWaveSampler
+...
+>>> sampler = DWaveSampler()
+>>> sampler.properties["default_readout_thermalization"]   # doctest: +SKIP
+0.0
+
+.. _property_drsim_default_shots:
+
+default_shots
+~~~~~~~~~~~~~
+
+Default setting for the number of times to run the circuit, as an integer.
+
+Example
+-------
+
+>>> from dwave.system import DWaveSampler
+...
+>>> sampler = DWaveSampler()
+>>> sampler.properties["default_readout_thermalization"]   # doctest: +SKIP
+0.0
+
+.. _property_drsim_default_time_limit:
+
+default_time_limit
+~~~~~~~~~~~~~~~~~~
+
+Default maximum runtime, in seconds, the solver is allowed to work on the
+given program, as a float.
+
+Example
+-------
+
+>>> from dwave.system import DWaveSampler
+...
+>>> sampler = DWaveSampler()
+>>> sampler.properties["default_readout_thermalization"]   # doctest: +SKIP
+0.0
+
+.. _property_drsim_default_transpile:
+
+default_transpile
+~~~~~~~~~~~~~~~~~
+
+Default setting, as a Boolean, for transpiling the submitted QCDL program.
+
+*   ``True``: Transpile the program.
+*   ``False``: Run the circuit exactly as specified in the submitted
+    QCDL or return an error
+
+Example
+-------
+
+>>> from dwave.system import DWaveSampler
+...
+>>> sampler = DWaveSampler()
+>>> sampler.properties["default_readout_thermalization"]   # doctest: +SKIP
+0.0
+
+.. _property_drsim_max_shots:
+
+max_shots
+~~~~~~~~~
+
+Maximum number of times the circuit can be executed, as an integer.
+
+Example
+-------
+
+>>> from dwave.system import DWaveSampler
+...
+>>> sampler = DWaveSampler()
+>>> sampler.properties["default_readout_thermalization"]   # doctest: +SKIP
+0.0
+
+.. _property_drsim_maximum_time_limit_s:
+
+maximum_time_limit_s
+~~~~~~~~~~~~~~~~~~~~
+
+Maximum time you can specify that your submitted circuit be executed, as a
+float.
+
+Example
+-------
+
+>>> from dwave.system import DWaveSampler
+...
+>>> sampler = DWaveSampler()
+>>> sampler.properties["default_readout_thermalization"]   # doctest: +SKIP
+0.0
+
+.. _property_drsim_minimum_time_limit_s:
+
+minimum_time_limit_s
+~~~~~~~~~~~~~~~~~~~~
+
+Minimum time you can specify that your submitted circuit be executed, as a
+float.
+
+Example
+-------
+
+>>> from dwave.system import DWaveSampler
+...
+>>> sampler = DWaveSampler()
+>>> sampler.properties["default_readout_thermalization"]   # doctest: +SKIP
+0.0
+
+.. _property_drsim_quota_conversion_rate:
+
+quota_conversion_rate
+~~~~~~~~~~~~~~~~~~~~~
+
+Rate at which user or project quota is consumed for the solver as a
+ratio to QPU solver usage. Different solver types may consume quota
+at different rates.
+
+Time is deducted from your quota according to:
+
+.. math::
+
+    \frac{num\_seconds}{quota\_conversion\_rate}
+
+See the :ref:`leap_hybrid_usage_charges` section for more information.
+
+Example
+-------
+
+>>> from dwave.system import DWaveSampler
+...
+>>> sampler = DWaveSampler()
+>>> sampler.properties["default_readout_thermalization"]   # doctest: +SKIP
+0.0
 
 .. list-table::
     :header-rows: 1
--- a/docs/workflow.rst
+++ b/docs/workflow.rst
@@ -1210,8 +1210,8 @@ Guidance on Mirroring
 
 .. _qcdl_simulator:
 
-Simulator
-=========
+QPU Simulator
+=============
 
 the |cloud|_ service provides a Monte Carlo simulator of QCDL programs. This is
 an ideal simulator built on top of Qiskit's
@@ -1235,11 +1235,13 @@ number of shots; however, its operation is
 
 The simulator in the |cloud|_ service supports two modes of simulations:
 
-*   Statevector Simulation (solver parameter `qpu=simulator`)
+*   Statevector Simulation (solver parameter :ref:`parameter_drsim_noise_model`
+    set to ``False``)
 
     This simulation is useful during initial testing of
     QCDL programs before introducing noise.
-*   Dual-Rail Erasure Simulation (solver parameter `qpu=DR17`)
+*   Dual-Rail Erasure Simulation (solver parameter
+    :ref:`parameter_drsim_noise_model` set to ``True``)
 
     This simulation is useful for exploring the impact of erasures on QCDL
     programs. It represents the quantum state as a Statevector with an array of
@@ -1255,16 +1257,13 @@ The following table compares these two simulation modes.
     *   -   Characteristic
         -   Statevector Simulation
         -   Dual-Rail Erasure Simulation
-    *   -   Run time
+    *   -   Runtime.
         -   Scales as :math:`O(2^n)` where :math:`n` is the number of qubits.
         -   Slightly slower than statevector simulation but same scaling.
-    *   -   Topology
-        -   No restrictions.
-        -   Depends on the noise model and the coupling map, options which model the selected QPU by default.
-    *   -   Depends on the noise model but which will represent the selected QPU by default.
+    *   -   Supported gates.
         -   All basis gates available in Qiskit (no transpilation require).
         -   Subset of basis gates (transpilation required).
-    *   -   Support for errors
+    *   -   Support for errors.
         -   No support.
         -   Supports the ``mced`` instruction to detect if the qubit has been
             erased, and the ``leak`` and ``seep`` instructions to simulate
@@ -1282,9 +1281,9 @@ dual-rail quantum computer by executing gate-model programs formulated as QCDL.
 The following documentation describes how to work with the |cloud|_ service:
 
 *   The :ref:`index_leap_sapi` section describes the |cloud|_ service.
-*   The :ref:`ocean_install` section explains how to install Ocean software.
 *   The :ref:`ocean_leap_authorization` section walks you through authorizing
     your Ocean client to access the simulator in the |cloud|_ service.
+*   The :ref:`ocean_install` section explains how to install Ocean software.
 
 Descriptions of the supported parameters and simulator properties are provided
 in the :ref:`qcdl_simulator_parameters` and :ref:`qcdl_simulator_properties`
@@ -1312,21 +1311,23 @@ The example below submits the following
 
     simulator_job_submission = bell_program()
 
-The program above is submitted to a dual-rail simulator, ``DR17``, in the
-|cloud|_ service.
+Submit the program above to a simulator for a dual-rail QPU with 17 qubits,
+``DRsim_17qubits``, in the |cloud|_ service.
 
 >>> from dwave.gate.qcdl.leap import LeapQCDLSimulator
 ...
 >>> simulator = LeapQCDLSimulator()         # doctest: +SKIP
 >>> future = simulator.run(                 # doctest: +SKIP
 ...     simulator_job_submission,
-...     qpu=`DR17`)
+...     qpu='DRsim_17qubits')
 >>> result = future.result().result         # doctest: +SKIP
+
+.. todo:: describe the results
 
 .. _qcdl_simulator_parameters:
 
 Simulator Parameters
---------------------
+====================
 
 The examples in this section submit the QCDL program defined in the
 :ref:`qcdl_submitting_programs_example` section.
@@ -1335,17 +1336,18 @@ The examples in this section submit the QCDL program defined in the
 .. _parameter_drsim_noise_model:
 
 noise_model
-~~~~~~~~~~~
+-----------
 
 Boolean flag that applies a noise model.
 
 *   ``noise_model=True``: Apply a noise model.
-*   ``noise_model=False``: Do not apply a noise model.
+*   ``noise_model=False``: Do not apply a noise model (simulate an ideal QPU,
+    as described in the :ref:`qcdl_simulator` section).
 
-Default is the value specified by the :ref:`property_drsim_default_noise_model`
+The default value is specified by the :ref:`property_drsim_default_noise_model`
 property.
 
-This example applies a noise model on the simulator.
+This example applies a noise model for the program submitted to the simulator.
 
 >>> from dwave.gate.qcdl.leap import LeapQCDLSimulator
 ...
@@ -1358,45 +1360,51 @@ This example applies a noise model on the simulator.
 .. _parameter_drsim_qpu:
 
 qpu
-~~~
+---
 
 The QPU to simulate, formatted as a string.
 
-Specified value must be one of the following values:
+The following values are supported:
 
-*   ``simulator``: Ideal simulator.
-*   ``DR17``: Dual-rail QPU with 17 qubits.
+*   ``DRsim_17qubits``: Dual-rail QPU with 17 qubits.
+*   ``DRsim_21qubits``: Dual-rail QPU with 21 qubits.
 
-Default is to simulate the QPU specified by the
+The default QPU to simulate is specified by the
 :ref:`property_drsim_default_qpu` property.
 
-This example selects the ``DR17`` QPU simulator.
+This example submits a QCDL program to a dual-rail QPU simulator with 17 qubits,
+``DRsim_17qubits`` .
 
 >>> from dwave.gate.qcdl.leap import LeapQCDLSimulator
 ...
 >>> simulator = LeapQCDLSimulator()         # doctest: +SKIP
 >>> future = simulator.run(                 # doctest: +SKIP
 ...     simulator_job_submission,
-...     qpu='DR17')
+...     qpu='DRsim_17qubits')
 >>> result = future.result().result         # doctest: +SKIP
 
 
 .. _parameter_drsim_repeat_until_shots_requested:
 
 repeat_until_shots_requested
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+----------------------------
 
 Boolean flag to run the circuit repeatedly until the requested number of
-measurements are accumulated.
+measurements, set by the :ref:`parameter_drsim_shots` parameter, are
+accumulated, even under noisy conditions.
 
 *   ``repeat_until_shots_requested=True``: Repeatedly run the circuit until
     the requested number of measurements, set by the
-    :ref:`parameter_drsim_shots` parameter, is accumulated.
+    :ref:`parameter_drsim_shots` parameter, is accumulated. Under noisy
+    conditions the circuit might be executed a greater number of times than set
+    by the :ref:`parameter_drsim_shots` parameter.
 *   ``repeat_until_shots_requested=False``: Run the circuit the number of times
-    set by the :ref:`parameter_drsim_shots` parameter.
+    set by the :ref:`parameter_drsim_shots` parameter. Under noisy conditions,
+    fewer measurements than requested by the :ref:`parameter_drsim_shots`
+    parameter might be accumulated.
 
-Default is to run the circuit the number of times set by the
-:ref:`parameter_drsim_shots` parameter.
+The default value is set by the
+:ref:`property_drsim_default_repeat_until_shots_requested` property.
 
 This example repeatedly executes the circuit to accumulate 10 measurements.
 
@@ -1412,7 +1420,7 @@ This example repeatedly executes the circuit to accumulate 10 measurements.
 .. _parameter_drsim_shots:
 
 shots
-~~~~~
+-----
 
 The number of times to run the circuit, formatted as an integer.
 
@@ -1435,7 +1443,7 @@ This example executes the circuit 15 times.
 .. _parameter_drsim_time_limit:
 
 time_limit
-~~~~~~~~~~
+----------
 
 Specifies the maximum runtime, in seconds, the solver is allowed to work on the
 given program. Can be a float or integer.
@@ -1460,7 +1468,7 @@ This example sets a maximum runtime of 20 seconds.
 .. _parameter_drsim_transpile:
 
 transpile
-~~~~~~~~~
+---------
 
 Boolean flag to rewrite the submitted QCDL circuit to use the QPU's supported
 basis gates and topology.
@@ -1487,19 +1495,16 @@ simulator.
 .. _qcdl_simulator_properties:
 
 Simulator Properties
---------------------
+====================
 
 .. _property_drsim_category:
 
 category
-~~~~~~~~
+--------
 
 Type of solver, as a string.
 
 *   ``software-gate``: Gate-model simulator.
-
-Example
--------
 
 >>> from dwave.gate.qcdl.leap import LeapQCDLSimulator
 ...
@@ -1510,15 +1515,12 @@ Example
 .. _property_drsim_default_noise_model:
 
 default_noise_model
-~~~~~~~~~~~~~~~~~~~
+-------------------
 
 Default setting for the application of a noise model, as a Boolean.
 
 *   ``True``: A noise model is applied.
 *   ``False``: Simulates an ideal QPU.
-
-Example
--------
 
 >>> from dwave.gate.qcdl.leap import LeapQCDLSimulator
 ...
@@ -1529,26 +1531,23 @@ False
 .. _property_drsim_default_qpu:
 
 default_qpu
-~~~~~~~~~~~
+-----------
 
 Default selection of the QPU to simulate, as a string.
 
-*   ``simulator``: Ideal simulator.
-*   ``DR17``: Dual-rail QPU with 17 qubits.
-
-Example
--------
+*   ``DRsim_17qubits``: Dual-rail QPU with 17 qubits.
+*   ``DRsim_21qubits``: Dual-rail QPU with 21 qubits.
 
 >>> from dwave.gate.qcdl.leap import LeapQCDLSimulator
 ...
 >>> simulator = LeapQCDLSimulator()         # doctest: +SKIP
 >>> simulator.properties["default_qpu"]     # doctest: +SKIP
-'simulator'
+'DRsim_21qubits'
 
 .. _property_drsim_default_repeat_until_shots_requested:
 
 default_repeat_until_shots_requested
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+------------------------------------
 
 Default setting, as a Boolean, for rerunning the circuit until the requested
 number of measurements is accumulated.
@@ -1556,9 +1555,6 @@ number of measurements is accumulated.
 *   ``True``: Repeatedly rerun the circuit.
 *   ``False``: Run the circuit the number of times set by the
     :ref:`parameter_drsim_shots` parameter.
-
-Example
--------
 
 >>> from dwave.gate.qcdl.leap import LeapQCDLSimulator
 ...
@@ -1569,12 +1565,9 @@ False
 .. _property_drsim_default_shots:
 
 default_shots
-~~~~~~~~~~~~~
+-------------
 
 Default setting for the number of times to run the circuit, as an integer.
-
-Example
--------
 
 >>> from dwave.gate.qcdl.leap import LeapQCDLSimulator
 ...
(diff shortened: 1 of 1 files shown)


docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+
+>>> from dwave.gate.qcdl.leap import LeapQCDLSimulator
+...
+>>> simulator = LeapQCDLSimulator()                 # doctest: +SKIP
+>>> simulator.properties["default_transpile"]       # doctest: +SKIP
+True
+
+.. _property_drsim_max_num_qubits:
+
+max_num_qubits
```
jstubbs-dwave, line 1629: This property is called `maximum_num_qubits`

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+
+>>> from dwave.gate.qcdl.leap import LeapQCDLSimulator
+...
+>>> simulator = LeapQCDLSimulator()         # doctest: +SKIP
+>>> simulator.properties["max_num_qubits"]       # doctest: +SKIP
+21
+
+.. _property_drsim_max_shots:
+
+max_shots
```
jstubbs-dwave, line 1645: This property is called `maximum_shots`.

jstubbs-dwave: We also provide a property for `minimum_shots`.

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+
+>>> from dwave.gate.qcdl.leap import LeapQCDLSimulator
+...
+>>> simulator = LeapQCDLSimulator()         # doctest: +SKIP
+>>> simulator.properties["default_shots"]   # doctest: +SKIP
+1
+
+.. _property_drsim_default_time_limit:
+
+default_time_limit
```
jstubbs-dwave, line 1597: This property is `default_time_limit_s`.

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+
+The number of measurements to run, formatted as an integer.
+
+Your QCDL program is executed once for each requested measurement.
+
+The specified value must not exceed the value of the
+:ref:`property_drsim_max_shots` property. Execution time is limited by the
+the value you specified in the :ref:`parameter_drsim_time_limit` parameter, or
+the maximum allowed runtime value of the
+:ref:`property_drsim_maximum_time_limit_s` property.
```
jstubbs-dwave, line 1440: I believe this should refer to the `default_time_limit_s` property, and not the maximum, since we apply the default if no `time_limit` parameter is provided.

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+    set by the :ref:`parameter_drsim_shots` parameter. Under noisy conditions,
+    fewer measurements than requested by the :ref:`parameter_drsim_shots`
+    parameter might be accumulated.
+
+The default value is set by the
+:ref:`property_drsim_default_repeat_until_shots_requested` property.
+
+If runtime exceeds the value you specified in the
+:ref:`parameter_drsim_time_limit` parameter, or the maximum allowed runtime
+value of the :ref:`property_drsim_maximum_time_limit_s` property, execution
```
jstubbs-dwave, line 1412: I believe this should refer to the default_time_limit_s property, and not the maximum, since we apply the default if no time_limit parameter is provided.

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
-    *   -   Run time
+    *   -   Runtime.
         -   Scales as :math:`O(2^n)` where :math:`n` is the number of qubits.
         -   Slightly slower than statevector simulation but same scaling.
-    *   -   Topology
-        -   No restrictions.
-        -   Depends on the noise model and the coupling map, options which model the selected QPU by default.
-    *   -   Depends on the noise model but which will represent the selected QPU by default.
+    *   -   Supported gates.
         -   All basis gates available in Qiskit (no transpilation require).
```
jstubbs-dwave, line 1264: The sentence in the parenthesis may be a typo, suggested change: "no transpilation required"

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+>>> simulator = LeapQCDLSimulator()         # doctest: +SKIP
+>>> future = simulator.run(                 # doctest: +SKIP
+...     simulator_job_submission,
+...     transpile=False)
+>>> result = future.result().result         # doctest: +SKIP
+
+
+.. _qcdl_simulator_properties:
+
+Simulator Properties
```
jstubbs-dwave, line 1510: We expose one additional property, `supported_qpu_strings`, containing a list of supported QPU strings, i.e., `["DRsim_17qubits", "DRsim_21qubits"]` which is not currently documented here.

jstubbs-dwave (changes requested):
Left some feedback on mismatching names, etc. Here is an example of the version 3 /solvers/remote/dr_simulator_<env> response:
```
{
  "identity": {
    "name": "dr_simulator_<env>"
  },
  "status": "ONLINE",
  "description": "Gate model simulator  - <ENV>",
  "properties": {
    "minimum_shots": 1,
    "maximum_shots": 1000000,
    "default_shots": 1000,
    "maximum_num_qubits": 21,
    "minimum_time_limit_s": 1,
    "maximum_time_limit_s": 2700,
    "default_time_limit_s": 2700,
    "default_noise_model": false,
    "supported_qpu_strings": [
      "DRsim_17qubits",
      "DRsim_21qubits"
    ],
    "default_qpu": "DRsim_21qubits",
    "default_transpile": true,
    "default_repeat_until_shots_requested": false,
    "parameters": {
      "time_limit": "Maximum runtime, in seconds, requested for the problem submission.",
      "shots": "Number of measurements to run.",
      "noise_model": "Apply a noise model to the simulation.",
      "transpile": "Transpile the circuit for the target QPU.",
      "repeat_until_shots_requested": "Repeatedly run the circuit to accumulate the required post-selected measurements.",
      "qpu": "Target gate-model QPU to simulate."
    },
    "supported_problem_types": [
      "qcdl"
    ],
    "version": "0.1",
    "category": "software-gate",
    "quota_conversion_rate": 1
  },
  "avg_load": 0.0
}
```

docs/workflow.rst
qci-amos: It's worth pointing out that this specifically is the text I want to eliminate here: I want both modes to support the boolean array.

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
     *   -   Characteristic
         -   Statevector Simulation
         -   Dual-Rail Erasure Simulation
-    *   -   Run time
+    *   -   Runtime.
         -   Scales as :math:`O(2^n)` where :math:`n` is the number of qubits.
         -   Slightly slower than statevector simulation but same scaling.
-    *   -   Topology
```
qci-amos, line 1304: Can we remove the comment about "slightly slower" or at least re-word it? The technical considerations aren't the same as when I originally wrote that line, e.g., simulator options that we exposed at the time I wrote that line aren't available currently.

Also, it's generally obviously true but worth calling out that `noise_model=False` will be faster than `noise_model=True`.

docs/workflow.rst
qci-amos: Also, it's not quite correct to say "all basis gates in Qiskit" because "basis gates" is by definition a subset of all possible gates. Maybe it's sufficient to just remove the word "basis" from this line.

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
-        -   No restrictions.
-        -   Depends on the noise model and the coupling map, options which model the selected QPU by default.
-    *   -   Depends on the noise model but which will represent the selected QPU by default.
+    *   -   Supported gates.
         -   All basis gates available in Qiskit (no transpilation require).
         -   Subset of basis gates (transpilation required).
```
qci-amos: It's not a subset of basis gates -- by definition "basis gates" is already the "minimally sufficient set". It definitely supports all of the basis gates, `rz`, `cz` (on edges in the coupling map), `sx`, and we've thrown in support for a handful more.

The phrase "transpilation required" could possibly be clarified because it's not required that WE do the transpilation.

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
 
-..  todo:: Update the rest once I can test in Leap
+.. _qcdl_submitting_programs:
+
+Submitting Programs
+===================
 
-.. list-table::
-    :header-rows: 1
+The gate-model simulator in the |cloud|_ service is intended to simulate the
```
qci-amos: The rendered documentation for `|cloud|_` isn't looking nice and doesn't go anywhere when clicked, perhaps it's missing its target?

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+dual-rail quantum computer by executing gate-model programs formulated as QCDL.
```
qci-amos: Lines like these are the reason I wanted to change how the `noise_model=False` was defined in the simulator because the current implementation has nothing to do with dual rail. Inconsistencies like these will be confusing, but there are a couple of ways we could close such gaps.

docs/workflow.rst
```diff
@@ -1278,11 +1235,13 @@ number of shots; however, its operation is
...
 
-*   Statevector Simulation
+*   Statevector Simulation (solver parameter :ref:`parameter_drsim_noise_model`
+    set to ``False``)
 
     This simulation is useful during initial testing of
     QCDL programs before introducing noise.
-*   Dual-Rail Erasure Simulation
+*   Dual-Rail Erasure Simulation (solver parameter
+    :ref:`parameter_drsim_noise_model` set to ``True``)
```
qci-amos: Could we move these 2 bullets to the table? I think it could be more explicitly clear that simulator features and limitations are controlled by this one boolean.

docs/workflow.rst
```diff
@@ -1278,11 +1235,13 @@ number of shots; however, its operation is
```
qci-amos, line 1227: The current simulator doesn't expose the `use_registers` simulator option, so this tip should remove mention of that.

docs/workflow.rst
```diff
@@ -1207,57 +1207,14 @@ Guidance on Mirroring
...
-.. compilation unsupported currently
 
-    You can simulate either the QCDL itself directly or a compiled version of the
-    QCDL (a ``.jmz`` file).
-
-.. todo:: update for Ocean
+.. _qcdl_simulator:
 
-.. _qcdl_submitting_simulator:
+QPU Simulator
```
qci-amos, line 1208: As you know, I've proposed changes in what happens on the backend. My suggestion is to choose wording here that is minimally sensitive to that sort of change.

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
-    *   -   Depends on the noise model but which will represent the selected QPU by default.
+    *   -   Supported gates.
         -   All basis gates available in Qiskit (no transpilation require).
         -   Subset of basis gates (transpilation required).
-    *   -   Support for errors
+    *   -   Support for errors.
         -   No support.
         -   Supports the ``mced`` instruction to detect if the qubit has been
             erased, and the ``leak`` and ``seep`` instructions to simulate
             leakage and seepage errors.
```
qci-amos, line 1268: Based on a DM he sent me, @qci-petrenko should review whether the 2nd phrase in this bullet point remains or should be removed.

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+
+Submit the program above to a simulator for a dual-rail QPU with 17 qubits,
+``DRsim_17qubits``, in the |cloud|_ service.
+
+>>> from dwave.gate.qcdl.leap import LeapQCDLSimulator
+...
+>>> simulator = LeapQCDLSimulator()         # doctest: +SKIP
+>>> future = simulator.run(                 # doctest: +SKIP
+...     simulator_job_submission,
+...     qpu='DRsim_17qubits')
```
qci-amos, line 1322: It's aesthetic... but the fact that this isn't the standard Python formatting (e.g., `black`) looks like a bug at a quick glance.

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+
+qpu
+---
+
+The QPU to simulate, formatted as a string.
+
+The following values are supported:
+
+*   ``DRsim_17qubits``: Dual-rail QPU with 17 qubits.
+*   ``DRsim_21qubits``: Dual-rail QPU with 21 qubits.
```
qci-amos: It is not explained to users what reason they'd have for choosing a 17 qubit qpu. I propose removing this option until there's a meaningful tradeoff for the user to consider.

Furthermore, whenever we do advertise these, we should present the respective coupling maps as images. We could consider hard-coding the coupling maps into this repo so that users have authoritative data they could feed into their own transpiler processes.

JoelPasvolsky:
> Left some feedback on mismatching names, etc. Here is an example of the version 3 /solvers/remote/dr_simulator_ response:

Thank you, @jstubbs-dwave, we'll need to ensure Leap is also updated because I took the names (e.g., ``max_shots``) from SAPI

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+
+.. _parameter_drsim_repeat_until_shots_requested:
+
+repeat_until_shots_requested
+----------------------------
+
+Boolean flag to run the circuit repeatedly until the requested number of
+measurements, set by the :ref:`parameter_drsim_shots` parameter, are
+accumulated, even under noisy conditions, where the accumulated measurements do
+not include erasures (see the :ref:`qcdl_basic_result_records` section).
```
qci-amos: This wording is a bit confusing. We already return all data including splats.

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+
+This example repeatedly executes the circuit to accumulate 10 measurements.
+
+>>> from dwave.gate.qcdl.leap import LeapQCDLSimulator
+...
+>>> simulator = LeapQCDLSimulator()         # doctest: +SKIP
+>>> future = simulator.run(                 # doctest: +SKIP
+...     simulator_job_submission,
+...     shots=10,
+...     repeat_until_shots_requested=True)
```
qci-amos, line 1437:
```suggestion
        noise_model=True,
...     repeat_until_shots_requested=True)
```

because otherwise `repeat_until_shots_requested` is meaningless

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+>>> result = future.result().result         # doctest: +SKIP
```
qci-amos, line 1438: perhaps we could a line showing somehow that `len(result.get_counts(post_select=True)[0] == 10`?

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+
+
+.. _parameter_drsim_repeat_until_shots_requested:
+
+repeat_until_shots_requested
+----------------------------
+
+Boolean flag to run the circuit repeatedly until the requested number of
+measurements, set by the :ref:`parameter_drsim_shots` parameter, are
```
qci-amos: I'm a bit uncomfortable with how the word "measurements" is used because nominally, each `measure(q)` is itself "a measurement", regardless of whether the outcome was a 0, 1, or *. Furthermore, there may be multiple such measurements per qubit per shot, and obviously in most cases, more than one qubit is measured.

If you're going to use a more abstract definition of "measurements" than what gate model users would expect, then it would be good to define what you mean.

I think it's important to define/use the term "yield"

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+
+*   ``repeat_until_shots_requested=True``: Repeatedly run the circuit until
+    the requested number of measurements, set by the
+    :ref:`parameter_drsim_shots` parameter, is accumulated. Under noisy
+    conditions the circuit might be executed a greater number of times than set
+    by the :ref:`parameter_drsim_shots` parameter.
+*   ``repeat_until_shots_requested=False``: Run the circuit the number of times
+    set by the :ref:`parameter_drsim_shots` parameter. Under noisy conditions,
+    fewer measurements than requested by the :ref:`parameter_drsim_shots`
+    parameter might be accumulated.
```
qci-amos:
```suggestion
    set by the :ref:`parameter_drsim_shots` parameter. Under noisy conditions,
    fewer measurements than requested by the :ref:`parameter_drsim_shots`
    parameter might be accumulated if `post_select=True`.
```

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+Boolean flag to run the circuit repeatedly until the requested number of
+measurements, set by the :ref:`parameter_drsim_shots` parameter, are
+accumulated, even under noisy conditions, where the accumulated measurements do
+not include erasures (see the :ref:`qcdl_basic_result_records` section).
+
+*   ``repeat_until_shots_requested=True``: Repeatedly run the circuit until
+    the requested number of measurements, set by the
+    :ref:`parameter_drsim_shots` parameter, is accumulated. Under noisy
+    conditions the circuit might be executed a greater number of times than set
+    by the :ref:`parameter_drsim_shots` parameter.
```
qci-amos: I think it's worth pointing out that the runtime is proportional to 1/yield and calling attention to the fact that runtime can grow exponentially.

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+    parameter might be accumulated.
+
+The default value is set by the
+:ref:`property_drsim_default_repeat_until_shots_requested` property.
+
+If runtime exceeds the value you specified in the
+:ref:`parameter_drsim_time_limit` parameter, or the maximum allowed runtime
+value of the :ref:`property_drsim_maximum_time_limit_s` property, execution
+terminates. The maximum number of times the circuit is run cannot exceed the
+value specified by the :ref:`property_drsim_max_shots` property.
```
qci-amos: This isn't true. Today max shots is only enforced on the `shots` input parameter, and not as an upper limit checked by the runtime when RUSR=true.

docs/workflow.rst
qci-amos: This wording isn't quite right because:
* `maximum_time_limit_s` is a validation of `time_limit` and
* `time_limit` is what's monitored at runtime

so once a `time_limit` is chosen, it's irrelevant to the simulation what max time is.

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+shots
+-----
+
+The number of measurements to run, formatted as an integer.
+
+Your QCDL program is executed once for each requested measurement.
+
+The specified value must not exceed the value of the
+:ref:`property_drsim_max_shots` property. Execution time is limited by the
+the value you specified in the :ref:`parameter_drsim_time_limit` parameter, or
```
qci-amos:
```suggestion
value you specified in the :ref:`parameter_drsim_time_limit` parameter, or
```

docs/workflow.rst
qci-amos: I have the same comment here about the relevance of max time in this context

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+:ref:`property_drsim_default_shots` property.
+
+This example executes the circuit 15 times.
+
+>>> from dwave.gate.qcdl.leap import LeapQCDLSimulator
+...
+>>> simulator = LeapQCDLSimulator()         # doctest: +SKIP
+>>> future = simulator.run(                 # doctest: +SKIP
+...     simulator_job_submission,
+...     shots=15)
```
qci-amos: if we're going to show the user how to override it, then might be good to anchor people with a reasonable number to choose instead of the default, e.g., "500" or "2000"

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+given program. Can be a float or integer.
+
+The specified time must be between the values of the
+:ref:`property_drsim_maximum_time_limit_s` and
+:ref:`property_drsim_minimum_time_limit_s` properties.
+
+The default runtime limit is specified by the
+:ref:`property_drsim_default_time_limit` property.
+
+This example sets a maximum runtime of 20 seconds.
```
qci-amos: I don't think 20 seconds is a reasonable time limit users should choose, even if it is above the minimum allowed time limit. Maybe 10 minutes and emphasize the units by showing `time_limit=10*60`?

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+This example requires that the QCDL circuit be submitted as written to the
+simulator.
+
+>>> from dwave.gate.qcdl.leap import LeapQCDLSimulator
+...
+>>> simulator = LeapQCDLSimulator()         # doctest: +SKIP
+>>> future = simulator.run(                 # doctest: +SKIP
+...     simulator_job_submission,
+...     transpile=False)
+>>> result = future.result().result         # doctest: +SKIP
```
qci-amos, line 1524: This isn't a meaningful example because if there's no noise, which is the default, then the results will not depend on the `transpile` option.

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+.. _property_drsim_default_noise_model:
+
+default_noise_model
+-------------------
+
+Default setting for the application of a noise model, as a Boolean.
+
+*   ``True``: A noise model is applied.
+*   ``False``: Simulates an ideal QPU, as described in the
+    :ref:`qcdl_simulator` section.
```
qci-amos, line 1556: Why not just link to the `noise_model` section instead of re-explaining what the values mean? Or consolidate the text describing the default with the text describing the option?

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+>>> simulator = LeapQCDLSimulator()         # doctest: +SKIP
+>>> simulator.properties["max_num_qubits"]       # doctest: +SKIP
+21
+
+.. _property_drsim_max_shots:
+
+max_shots
+---------
+
+Maximum number of times the circuit can be executed, as an integer.
```
qci-amos: As I mentioned elsewhere, this isn't enforced by the simulator (or even provided to the simulator). Hence if RUSR=True then this statement isn't true.

docs/workflow.rst
```diff
@@ -1298,51 +1257,485 @@ The following table compares these two simulation modes.
...
+
+maximum_time_limit_s
+--------------------
+
+Maximum time, in seconds as a float, that your submitted circuit can run.
+
+This value limits the range of values you can set on the
+:ref:`parameter_drsim_time_limit` parameter and also limits the runtime allowed
+for a program submitted with the
+:ref:`parameter_drsim_repeat_until_shots_requested` set to true.
```
qci-amos:
```suggestion
This value limits the range of values you can set on the
:ref:`parameter_drsim_time_limit` parameter.
```

docs/workflow.rst
randomir: This is standard Python (PEP-8) formatting.

docs/workflow.rst
qci-amos: The statement "introduce error modeling in a controlled way" is exactly what users will want but will not be supported with this release.

docs/workflow.rst
randomir: Coupling map should be available as part of solver metadata (properties) on SAPI.

docs/workflow.rst
qci-amos: Also, they're both statevector based, so it would be confusing to see this sentence worded in a way that makes it seem like it's a difference between the two. If you want to preserve the word, then qualify it with something like "pure".

docs/workflow.rst
JoelPasvolsky: Perhaps not now but I would like to see this property renamed to something like `supported_qpu_names`, `simulated_models`, `simulated_systems`, `supported_simulations` or something that does not put the format of the property's value in its name

docs/workflow.rst
JoelPasvolsky: Sure.

Image

docs/workflow.rst
JoelPasvolsky: Thanks!

Image

docs/workflow.rst
JoelPasvolsky: That substitution is defined in the [SDK conf.py](https://github.com/dwavesystems/dwave-ocean-sdk/blob/master/docs/conf.py) file. Not defining it in the [local conf.py](https://github.com/dwavesystems/dwave-gate/blob/main/docs/conf.py) file generates HTML with that look and also a bunch of warnings, but having it defined in both places causes a warning in the SDK build.

Last time I checked I couldn't see a way of having both without warnings but when I have an hour I'll try again.

docs/workflow.rst
JoelPasvolsky:
```
The :ref:`QPU simulator <qcdl_simulator>` in the |cloud|_ service is intended to
simulate :ref:`gate-model quantum computers <qpu_gate_model_intro>` by executing
programs formulated as QCDL.
```

Image

docs/workflow.rst
jstubbs-dwave: I will add a ticket to the backlog and ensure we coordinate a documentation update when this is done.

JoelPasvolsky: Thank you, @jstubbs-dwave

docs/workflow.rst
```diff
@@ -1298,51 +1257,510 @@ The following table compares these two simulation modes.
...
+
+Default setting for the number of measurements to run (times to execute your
+QCDL circuit), as an integer. With dual-rail QPUs, a measurement result can be a
+"splat" (see the :ref:`qcdl_basic_measurements` section).
+
+>>> from dwave.gate.qcdl.leap import LeapQCDLSimulator
+...
+>>> simulator = LeapQCDLSimulator()         # doctest: +SKIP
+>>> simulator.properties["default_shots"]   # doctest: +SKIP
+1
```
jstubbs-dwave: `default_shots` is now `1000`.

jstubbs-dwave (changes requested):

docs/workflow.rst
JoelPasvolsky: @qci-amos, for your 4 comments on this section: 

Image

docs/workflow.rst
JoelPasvolsky: Added ``print(sum(result.get_counts(post_select=True)[0].values())) # doctest: +SKIP``

Image

docs/workflow.rst
JoelPasvolsky: We prefer that a reference section not send users to another section, and try to do that only when there's a high maintenance cost to info duplication. In this case the meaning of True/False options won't change and the content is short, so low risk of needing to update and more user friendly. 
The description of the simulator has high maintenance and high likelihood of updates so that info is not duplicated here and instead a link is given.

jstubbs-dwave (approved):

docs/workflow.rst
qci-amos, line 1: We made a recent change in the simulator, and this note:
> A barrier() instruction does not necessarily imply a sync() (see the Synchronization section).

is no longer true: barrier and sync are now synonyms in the simulator (and we'll update the compiler soon)

docs/workflow.rst
```diff
@@ -1207,13 +1207,552 @@ Guidance on Mirroring
...
 
+
+.. _qcdl_simulator:
+
+QPU Simulator
+=============
+
+the |cloud|_ service provides a Monte Carlo simulator of QCDL programs. This is
+an ideal simulator built on top of Qiskit's
+`AerStatevector <https://qiskit.github.io/qiskit-aer/stubs/qiskit_aer.quantum_info.AerStatevector.html>`_.
```
qci-amos, line 1213:
```suggestion
the |cloud|_ service provides a Monte Carlo simulator of QCDL programs. This is
built on top of Qiskit's
`AerStatevector <https://qiskit.github.io/qiskit-aer/stubs/qiskit_aer.quantum_info.AerStatevector.html>`_.
```

I'm not sure what "ideal" means here?

docs/workflow.rst
```diff
@@ -1207,13 +1207,552 @@ Guidance on Mirroring
...
+classical and quantum components of the hardware and then executes your QCDL
+instructions one at a time to update that state. As a Monte Carlo simulator, it
+is significantly slower than a "sampling" simulator and scales linearly with the
+number of shots; however, its operation is
+`embarrassingly parallel <https://en.m.wikipedia.org/wiki/Embarrassingly_parallel>`_.
+
+.. tip::
+    Accuracy bears a simulation cost and error handling increases circuit
+    complexity. It is advisable to start circuit development against the ideal
+    simulator and then introduce error modeling.
```
qci-amos, line 1226: I like how you added "for future consideration". Can you add another note to remind us to update this paragraph as new features come available?

docs/workflow.rst
```diff
@@ -1207,13 +1207,552 @@ Guidance on Mirroring
...
+    *   -   Runtime.
+        -   Scales as :math:`O(2^n)` where :math:`n` is the number of qubits.
+
+            Faster.
+        -   Slower but same scaling.
+    *   -   Supported gates.
+        -   All gates available in Qiskit (no transpilation required).
+        -   Subset of gates (transpilation required).
+    *   -   Support for errors.
+        -   No support.
```
qci-amos: you can mention that it supports `mced`, but that it unconditionally returns `0` (i.e., "not leaked").

docs/workflow.rst
```diff
@@ -1207,13 +1207,552 @@ Guidance on Mirroring
...
+        -   Solver parameter :ref:`parameter_drsim_noise_model` set to
+            ``True``.
+
+            This simulation is useful for exploring the impact of erasures on
+            QCDL programs. It operates by randomly applying Pauli errors,
+            leakages, and seepages after quantum gates and idles.
+    *   -   Runtime.
+        -   Scales as :math:`O(2^n)` where :math:`n` is the number of qubits.
+
+            Faster.
```
qci-amos:
```suggestion
```

the other one already says "slower" and... as you know my standing request is to change "statevector simulation" to look more like "DR simulation" in which case there will be less of a performance difference so I'd prefer to minimize our descriptions of the differences.

docs/workflow.rst
```diff
@@ -1207,13 +1207,552 @@ Guidance on Mirroring
...
+            Useful during initial testing of QCDL programs before introducing
+            noise.
+        -   Solver parameter :ref:`parameter_drsim_noise_model` set to
+            ``True``.
+
+            This simulation is useful for exploring the impact of erasures on
+            QCDL programs. It operates by randomly applying Pauli errors,
+            leakages, and seepages after quantum gates and idles.
+    *   -   Runtime.
+        -   Scales as :math:`O(2^n)` where :math:`n` is the number of qubits.
```
qci-amos: I think we should generalize this to say `O(S · g · 2ⁿ)` where `S` is the number of shots and `g` is the number of gates executed. I think that would be more clear to a user what's happening, and it also helps facilitate the discussion for if/when we add a sampling based simulator which would be roughly `O(g · 2ⁿ + S·n)`

docs/workflow.rst
qci-amos: [PEP-8](https://peps.python.org/pep-0008/#indentation) is explicitly permissive in this regard:

> The closing brace/bracket/parenthesis on multiline constructs may either line up under the first non-whitespace character of the last line of list, as in:
> ...
> or it may be lined up under the first character of the line that starts the multiline construct, as in:


`black` and `ruff` are similar to each other and are stricter than PEP-8 (and imho a lot easier to read to me and diffs are cleaner)

I asked Claude if this is controversial and it says:

> The broader thing that defused all of this is that most teams decided arguing about it was worse than any particular answer. Black's pitch was never that its style is optimal, just that it's uniform and nobody has to litigate it in review. Since PEP 8 explicitly permits all three closing-bracket positions, there was no authority to appeal to — a tool had to pick, and picking is what made the question stop mattering.
>
> If you're on a codebase without a formatter, your original snippet is fine and flake8 will pass it. If you're adding one, expect it to be rewritten, and don't read that as a judgment on the original.

docs/workflow.rst
```diff
@@ -1207,13 +1207,552 @@ Guidance on Mirroring
...
+
+Running a circuit might return, under noisy conditions, measurements that are
+declared to be “erased”, as described in the :ref:`qcdl_basic_measurements`
+section. To try to achieve the required number of non-erasure measurements
+indicted by the :ref:`parameter_drsim_shots` parameter, as counted after
+post-selection to remove "splats" (see the :ref:`qcdl_basic_result_records`
+section), you can select to repeatedly run the circuit. With yield defined as
+the percentage of shots without erasures, the required number of executions (and
+runtime) is proportional to the value of the :ref:`parameter_drsim_shots`
+parameter and the reciprocal of the yield, and can grow exponentially with
```
qci-amos:
```suggestion
parameter and the reciprocal of the yield, and will grow exponentially with
```

I mean if you're on the lower side of the exponential curve you might not notice, but technically it's exponential none the less.

The only way it's not exponential is if the user has implemented error correction, but I'm not sure how many qualifications we want to give to all the descriptions and you already said "exponentially with noise".

docs/workflow.rst
```diff
@@ -1207,13 +1207,552 @@ Guidance on Mirroring
...
+increased noise.
+
+*   ``repeat_until_shots_requested=True``: Repeatedly run the circuit until
+    the requested number of non-erasure measurements, indicated by the
+    :ref:`parameter_drsim_shots` parameter, is accumulated with
+    ``post_select=True``. Under noisy conditions the circuit might be executed a
```
qci-amos, line 1410: can you link to the `get_counts` api here?

docs/workflow.rst
qci-amos: that looks great.

would it be possible to add a second `print` showing that the number is (non-deterministic) but higher if `post_select=False`?

docs/workflow.rst
```diff
@@ -1207,13 +1207,552 @@ Guidance on Mirroring
...
+Submit the program above to a simulator for a dual-rail QPU with 17 qubits,
+``DRsim_17qubits``, in the |cloud|_ service.
+
+>>> from dwave.gate.leap import LeapQCDLSimulator
+...
+>>> simulator = LeapQCDLSimulator()         # doctest: +SKIP
+>>> future = simulator.run(                 # doctest: +SKIP
+...     simulator_job_submission,
+...     qpu='DRsim_17qubits')
+>>> result = future.result().result         # doctest: +SKIP
```
qci-amos, line 1323: what do you think about adding an example `print` similar to RUSR above just showing users the shape of the returned data? I mean you don't need to add a `print` to every example, so perhaps just the leading example here?

docs/workflow.rst
qci-amos: Why are default values handled this way and not in the client-facing API? I'm guessing it's because in other systems perhaps there are multiple clients so you want to control in a centralized way?

If that's the reason, currently we have only one client, i.e., `LeapQCDLSimulator` but even if/when we start exposing the simulator locally we could handle default values in a base class. Or perhaps the reason is that Leap is expected to handle default values and not Ocean?

Anyway, it seems like we won't have flexibility here and certainly not in the short term, but I'm just pointing out that this way of showing parameters and their default values seems less clear to me than a docstring.

docs/workflow.rst
JoelPasvolsky: I have a [branch](https://github.com/JoelPasvolsky/dwave-gate/tree/results) in progress to add a `Results` section to the QCDL API and then yes

docs/workflow.rst
qci-amos, line 1: reminder: there are several remaining references to `Aqumen` in this file

docs/workflow.rst
qci-amos: This is worth consideration because we do not currently have `leak` or `seep` in `operations.py`. We should either add them there or remove mention of them here.

My view is that the solution most consistent with requirements is to add them to `operations.py`.

docs/workflow.rst
JoelPasvolsky: Ideally, we'll add a mock ``LeapQCDLSimulator`` class that will allow us to remove the  ``# doctest: +SKIP`` directives, and then I'd feel much more comfortable increasing the complexity of these examples. While we're not yet validating examples through doctest, I worry that if code changes we could get erroneous examples that go undetected.
How strongly do you feel about adding such a line of code here, @qci-amos?

docs/workflow.rst
JoelPasvolsky: These parameters & properties are coming in from SAPI, not the Ocean client.

docs/workflow.rst
JoelPasvolsky: removed

docs/workflow.rst
JoelPasvolsky: Updated to "... and grows..." (present tense)

docs/workflow.rst
JoelPasvolsky: Added

Image

docs/workflow.rst
qci-amos: not strongly, I'm happy to go with your recommendation

qci-amos (approved):
I'd still like @qci-petrenko to review this document, but in the interest of getting changes through asap and not blocking Joel I'm approving since we can refine things later.

PR #81, opened 2026-09-08
Update package and file structure

thisac:
All renames and moves were done using `git mv`, except for the extraction for the `display` module.

* [Remove qcdl_ prefix from modules](https://github.com/dwavesystems/dwave-gate/commit/e4983574b6607bca8c8280c85c2aeab49eb172b0)

* [Move leap.py and implementations.py outside qcdl module.](https://github.com/dwavesystems/dwave-gate/commit/a4c23a979ba63d97ba90391abca39f6bb54c2379)

* [Hide QCDL internal utils](https://github.com/dwavesystems/dwave-gate/commit/f6ce17ff7e5d4e6eb79749a772e6ff37f6b907c5)

* [Move display helpers to dwave.gate.utils.display](https://github.com/dwavesystems/dwave-gate/commit/22b7567a4f88199d9f46e653202158fdb57c0fb1)

### AI disclosure
Claude was used to identify changes and review the file renames and moves.


dwave/gate/qcdl/__init__.py
```diff
@@ -16,12 +16,11 @@
 from .components import QCDLModule, QCDLModuleContainer, Scope, procedure
 from .constants import LogicalOutcomeToInteger
 from .exceptions import QCDLInternalError, QCDLUserError
-from .leap import LeapQCDLSimulator
-from .qcdl_circuit import qcdl
-from .qcdl_models import QCDLProgram, QCDLProcedureDef, QCDLStatement
+from ..leap import LeapQCDLSimulator
```
randomir: If `leap` is outside of `qcdl` now, why keep this here?
```suggestion
```

dwave/gate/qcdl/operations.py
```diff
@@ -60,7 +60,7 @@ def qcdl_module_methods(q0, q1):
 
 import numpy as np
 
-from . import implementations
+from .. import implementations
```
randomir, line 62: Similarly to leap, do we still need this here? Feels a bit like code smell.

randomir (approved):
LGTM!

Very nice readability improvements.

dwave/gate/qcdl/__init__.py
thisac: Good catch. Removing.

dwave/gate/qcdl/operations.py
thisac: Yeah, it's a bit smelly since they're used inside `measure` and `mced`. `implementations.py` does feel quite stand-alone though, and like something a user might look for _outside_ of the QCDL implementation scope. :thinking:

thisac: I'd probably just leave it like this, unless `mirror_measurement_register()` and `mirror_bool_register()` are explicitly only for use in `operations.py`, but they seem to me to be more general in design and created for use as stand-alone user-facing procedures.

qci-amos: They might be useful on their own. Perhaps the name `implementations` isn't quite right? The name originated from another, similar file elsewhere and in my mind, we could add logic to it to serve both as utilities for users but also as working examples they could build their own logic off of. On one hand, we could just move them to `operations.py` however, I think I'd prefer to keep a bit of separation but also a kind of sibling relationship between classical and quantum logic. Maybe it could be renamed `algorithms.py`? or `helpers.py`? I'm not sure what's least smelly.

qci-amos: but if we start adding logic like simulator and other non-qcdl features to this repo, then I feel like "implementations" will more naturally fit close to operations.py

qci-amos (approved):
thank you! lgtm which ever way you want to go with `implementations.py`

PR #83, opened 2026-09-10
Add ``Results`` class and related functions to QCDL API

JoelPasvolsky:
CC @qci-petrenko, you can see the HTML rendered on the tests below

@qci-amos, needed for SDK release

### DIFF
--- a/dwave/gate/results.py
+++ b/dwave/gate/results.py
@@ -214,37 +214,38 @@ def count_measurements(
 ) -> dict[int | str, int]:
     """Convert memory to a dict of observation counts.
 
-    Raw data (memory) is just 0, 1, or ``"*"`` (called "splat") for each qubit,
-    for each shot. This method will convert that into a count of each set of
+    Raw data (memory) contains values :math:`0, 1` or ``"*"`` ("splat") for each
+    qubit, for each shot. This method converts that into a count of each set of
     measurements. The key for the dict is customizable.
 
-    When converting to int/bin, this treats the qubit register as big endian,
-    i.e., that the register has the least significant bit last in the array for
-    each shot.
+    When converting to int/bin, the qubit register is treated as big endian;
+    i.e., the least-significant bit is last in the array for each shot.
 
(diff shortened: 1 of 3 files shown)


thisac (approved):
LGTM

dwave/gate/results.py
```diff
@@ -312,25 +313,24 @@ def count_measurements(
...
-    fields do not cause validation failures.
+    """Result returned by a gate-model simulator or QPU.
 
     Attributes:
-        num_shots: Total number of shots executed.
+        num_shots: Total number of times the circuit was executed.
         start_time: Timestamp for when execution began.
         end_time: Timestamp for when execution ended.
-        seconds_per_shot: Average wall-clock time per shot, in seconds.
+        seconds_per_shot: Average per-shot wall-clock time, in seconds.
```
qci-amos: actually, per the Jira I saw, maybe we want the wording here to reflect that the circuit does parallelism? currently the calculation is total_wallclock_time / num shots, but that's not quite right if there are parallel processes.

perhaps AquSim should compute a number that reflects this, or we could update the wording here to say something more general like "total time / num shots"

dwave/gate/results.py
```diff
@@ -214,37 +214,38 @@ def count_measurements(
...
+    When converting to int/bin, the qubit register is treated as big endian;
+    i.e., the least-significant bit is last in the array for each shot.
 
     :data:`key_format` values:
-        * ``"bin"``: keys will be binary numbers with width taken from the memory
-        * ``"hex"``: keys will be hexadecimal strings (e.g., ``'0x3'``)
-        * ``None``: keys will be integers
+        * ``"bin"``: keys are binary numbers with width taken from the memory.
+        * ``"hex"``: keys are hexadecimal strings (e.g., ``'0x3'``).
+        * ``None``: keys are integers.
```
qci-amos, line 227: we should just remove these, but perhaps once my PRs get merged

qci-amos (approved):

PR #84, opened 2026-09-11
Add example results

JoelPasvolsky:


randomir (approved):
LGTM!

docs/workflow.rst
```diff
@@ -1311,23 +1311,49 @@ The example below submits the following
...
+For one particular execution of the program above, the following counts are
+returned.
+
+>>> print(result.get_counts())              # doctest: +SKIP
+[{'11': 203, '*1': 14, '00': 230, '0*': 22, '*0': 19, '1*': 10, '**': 2}]
+
+The execution time for that submission is about a tenth of a second.
+
+>>> print(result.run_time)                  # doctest: +SKIP
+0.093493
```
qci-amos, line 1353: it might be worth explaining that this is the time actually simulating and has nothing to do with time to warm up the instance etc

docs/workflow.rst
```diff
@@ -1311,23 +1311,49 @@ The example below submits the following
...
-(1, 1000, 2)
+(1, 500, 2)
+
+For one particular execution of the program above, the following counts are
+returned.
+
+>>> print(result.get_counts())              # doctest: +SKIP
+[{'11': 203, '*1': 14, '00': 230, '0*': 22, '*0': 19, '1*': 10, '**': 2}]
+
+The execution time for that submission is about a tenth of a second.
```
qci-amos, line 1350: if you're describing it as a "submission" then that sounds end-to-end to me and perhaps should reflect some things like queuing, instance warmup, python launching, etc

qci-amos (approved):

docs/workflow.rst
JoelPasvolsky: Thanks, updated to say "The execution time on the simulator (excluding any queuing time, for example)
for that job submission is about a tenth of a second."

PR #85, opened 2026-09-11
Onboarding beta testers

JoelPasvolsky:
@randomir, see if there's more needed or worth changing.


randomir (approved):
LGTM!

Minor note: onboarding content would probably make more sense in the  Ocean SDK docs, under (perhaps) gate-model page. But I guess it's fine to temporarily host it here. 🤷

docs/workflow.rst
```diff
@@ -4,13 +4,106 @@
...
 
+.. _qcdl_onboarding:
+
+Onboarding for Beta Testers
+===========================
+
+.. important::
+    Features for real-time control, which are being phased into dual-rail
+    quantum computing systems, are already available on the simulator in the
+    |cloud|_ service for prototyping and learning.
```
qci-amos, line 22: I'm glad to see this mentioned somewhere

docs/workflow.rst
```diff
@@ -4,13 +4,106 @@
...
+projects (see the "Multiple Leap Projects" tab).
+
+The following are two simple ways to use the beta-tester project's API token
+from your existing development environment.
+
+*   Add a section to your ``dwave.conf`` file.
+
+    You can see your ``dwave.conf`` file using the methods described in the
+    :ref:`cloud_configuration` or using the :ref:`D-Wave CLI <ocean_dwave_cli>`
+    section.
```
qci-amos, line 80: it would be nice if there was a cli way to do this

docs/workflow.rst
```diff
@@ -4,13 +4,106 @@
...
+
+        [beta]
+        token = BETA-123456789123456789123456789
+
+    You can then set ``profile="beta"`` to use the beta-tester project's API
+    token when accessing the simulator.
+
+    >>> from dwave.gate.leap import LeapQCDLSimulator
+    ...
+    >>> simulator = LeapQCDLSimulator(profile="beta")         # doctest: +SKIP
```
qci-amos, line 95: isn't there a way to configure a default in `dwave.conf`?

docs/workflow.rst
```diff
@@ -4,13 +4,106 @@
...
+    :ref:`cloud_configuration` or using the :ref:`D-Wave CLI <ocean_dwave_cli>`
+    section.
+
+    For example, add a ``beta`` section::
+
+        [defaults]
+        token = ABC-123456789123456789123456789
+
+        [beta]
+        token = BETA-123456789123456789123456789
```
qci-amos, line 88: my conf has `leap_client_id` and `metadata_api_endpoint`. Are these not needed for the beta testers?

docs/workflow.rst
```diff
@@ -4,13 +4,106 @@
...
+accept the emailed invitation to a new project. You use the API token from this
+project to access and send jobs to the simulator.
+
+The :ref:`ocean_leap_authorization` section describes how to work with multiple
+projects (see the "Multiple Leap Projects" tab).
+
+The following are two simple ways to use the beta-tester project's API token
+from your existing development environment.
+
+*   Add a section to your ``dwave.conf`` file.
```
qci-amos, line 76: is ti clear to users how/where to create this file? It hasn't been so for me and my team

qci-amos (approved):
i'm not really the right person to review all this, but the parts I am familiar with lgtm

docs/workflow.rst
randomir: `dwave config ls` should list all available `dwave.conf` files, in order of preference.

randomir: But you make a good point, @qci-amos, we should advise users explicitly to just run this:

```
dwave config create --profile beta
```

That command will locate the config file and/or create one if it doesn't exist; then it will create a new profile called `beta` (or update the existing one); and ask the user for the SAPI token.

docs/workflow.rst
randomir: Not needed. The PROD values for these params are shipped with the client. You had to manually override/specify them if you ran in non-PROD envs.

docs/workflow.rst
randomir: `dwave config create` will update the default profile _by default_. That makes sense for new D-Wave users that are trying this out, but not for existing users. Now, I don't know who's going to be onboarding here.

docs/workflow.rst
randomir: Perhaps we should provide a **fully automated flow** example.

1. Authorize Ocean to access Leap:
```
$ dwave auth login
Please visit the following URL to authorize Ocean: https://cloud.dwavesys.com/leap/openid/authorize?...
Authorization completed successfully. You can now use "dwave auth get" to fetch your token.
```

2. Configure your `beta` profile:
```
$ dwave config create --profile beta --auto-token --project "D-Wave Gate Model Simulator Beta"
Using the simplified configuration flow.
Try 'dwave config create --full' for more options.

Updating existing configuration file: /home/radomir/.config/dwave/dwave.conf
Creating new profile: beta
Fetched SAPI token for project 'D-Wave Gate Model Simulator Beta' (R4zH) from Leap API.
Configuration saved.
```

Or as a one-liner:
```
dwave auth login && dwave config create --profile beta --auto-token --project "D-Wave Gate Model Simulator Beta"
```

Note I used the full project name for clarity, but it can be shortened to `R4zH`:
```
dwave auth login && dwave config create --profile beta --auto-token --project R4zH
```

docs/workflow.rst
qci-amos: so "beta" is some sort of alias/subset of PROD?

docs/workflow.rst
qci-amos: most of the examples that Trevor K and I are planning, he wouldn't specify `profile=` so ideally we could have a way to guide users to simply configure their `dwave.conf` appropriately instead of either encoding the profile in the examples or asking users to do that

docs/workflow.rst
randomir: No, "beta" (token) here refers to a new project "D-Wave Gate Model Simulator Beta". It's a project in the PROD environment.

docs/workflow.rst
randomir: You can also specify the default profile in the `DWAVE_PROFILE` env var. That's the easiest way to "specialize" an environment. A shell, or a container.

```bash
$ DWAVE_PROFILE=beta python
>>> from dwave.gate.qcdl.leap import LeapQCDLSimulator
>>> simulator = LeapQCDLSimulator()
>>> simulator.properties
{'max_shots': 1000000, ...
```

docs/workflow.rst
JoelPasvolsky: Thank you, I added the following:

Image

docs/workflow.rst
```diff
@@ -4,13 +4,118 @@
...
+
+    You can use the following :ref:`D-Wave CLI <ocean_dwave_cli>` commands to
+    authorize Ocean software to access the |cloud|_ service and Configure a
+    ``beta`` profile,::
+
+        $ dwave auth login
+        $ dwave config create --profile beta --auto-token --project "Beta Testing"
+
+    where ``Beta Testing`` should be replaced with the project name as displayed
+    in the |cloud|_ service. The commands locate your existing configuration
```
randomir, line 105: Perhaps worth mentioning that once Ocean is authorized to access Leap (the first command above), you can list the available projects with:
```
$ dwave leap project ls
```

JoelPasvolsky: I think for this small number of users in a single overall use case it's okay to stick to two flavours: (1) they go from the welcome email to onboarding (2) they get the project from the Leap website. Any small inconvenience from not having that command at hand here for this one time use seems acceptable to me.

PR #90, opened 2026-09-15
Increment version to 0.6.0

thisac:
0.6.0
=====

New Features
------------

- ``QCDLModule.one_to_all`` now accepts any group of qubits as its ``destinations``: a ``Scope``, a single ``QCDLModule``, or a sequence of either. Qubits are deduplicated by name, preserving the order in which they are given.

- Validate the arguments of the ``@qcdl`` decorator and the signature of the function it decorates. Because qubits are injected by name, a parameter that is not a ``q<N>`` previously received nothing and Python's own error named only the parameter, giving no hint that qubits were involved, and a generated qubit with no parameter to receive it was dropped from the program silently. Both now raise and explain the ``q<N>`` rule and which qubits were supplied. ``num_qubits`` is checked when the decorator is applied, including a bare ``@qcdl`` that would otherwise be passed the decorated function as the qubit count.

Upgrade Notes
-------------

- The ``Result.get_memory`` and ``Result.get_counts`` methods now default to the measured register rather than to every qubit in the circuit, so bitstrings no longer include a padded ``unmeasured_value`` for a qubit that was never measured. Pass ``register=get_default_register(...)`` explicitly to keep the previous, padded register.

Bug Fixes
---------

- Fix ``QCDLModule.one_to_all`` raising an ``AttributeError`` when ``destinations`` is a sequence of qubits, such as the ``qcdl_modules`` of a ``Scope``, rather than the ``Scope`` itself.

- The ``Result.get_memory`` and ``Result.get_counts`` methods no longer pad unmeasured qubits into the bitstring by default, where such a qubit was previously read as an outcome. They default to ``Result.get_measurements_register`` and fall back to every qubit only when nothing was measured at all.

- The ``YieldHandling.apply`` method now accepts every key format that the ``count_measurements`` function produces. Testing keys for a splat raised ``TypeError`` on the integer keys produced for ``key_format=None``, and the all-erasures fallback key is now built only for string keys, which are the only ones that can hold a splat.

- The ``YieldHandling.apply`` method now raises ``ValueError`` for an empty distribution, rather than the ``StopIteration`` that its fallback key raised for a counts dict that post selection had emptied.


randomir (approved):
LGTM. But next time we should ask Claude to keep release notes short. These are novels nobody cares to read 😅

thisac:
> But next time we should ask Claude to keep release notes short. These are novels nobody cares to read

Sure, I agree they could be shorter. Add some new rules to [our agents](https://github.com/dwavesystems/dwave-ocean-sdk/pull/487)?

